PR #19, opened 2026-07-26
MCP: read-only stdio server for LLM apps to query workout history

Pengboi:
## What this is

Lets a self-hoster ask an AI assistant (Claude Desktop, Cursor, Cline, Continue — anything that speaks the [Model Context Protocol](https://modelcontextprotocol.io)) about their own workouts in their own words: *"what did I bench last week?", "what's my estimated 1RM on deadlift?", "which muscles have I been neglecting this month?"*. The LLM reads your `./data` directly — no new container, no extra auth on the api server, no third-party service.

The server is a tiny Node script the LLM client spawns locally as a stdio process (per the official MCP design — Claude Desktop, Cursor, Cline all work this way). It reads the same `state-<uid>.json` files your `api` container writes, reuses the exact pure-function training helpers from `frontend/src/lib/*.js` (so the numbers it answers with match what the Stats screen shows), and exits when the LLM client disconnects.

Phase 1 is read-only. A future phase would add write tools (`log_workout`, `bump_weight`, …) but that needs a long-lived token path that doesn't exist yet on the api side, plus a write-lock against the web UI's read-modify-write. Out of scope for this PR.

## Why someone might want it

openGym's pitch is "your data, your server". A self-hoster who already runs Claude Desktop or Cursor can ask their AI about their training in natural language *without* their data ever leaving the box where they already run docker. The LLM only sees what its host process can already see on disk — passkey keys, VAPID secrets, session cookies are never read.

If you don't use an AI assistant, this PR doesn't affect you in any way. The frontend is untouched; `api/server.js` is untouched; `docker-compose.yml` is untouched; the only new runtime dependency on the host is for someone who opts in by adding the server to their LLM client's config.

## What landed

Four commits, separated per the project's per-feature cadence:

1. **`i18n: split the pure core out so the lib is Node-loadable`** — The one lib file that wasn't Node-safe was `i18n.js` (uses Vite's `import.meta.glob` at module top level). Pulled the runtime-agnostic state, constants and readers into `i18n-core.js`; `i18n.js` keeps the Vite-only bits (async `setLang`, the React hook) and re-exports from core. None of the 23 existing import sites change; `format.js` and `exercises.js` import `i18n-core.js` directly. `exercises.js` got a one-line `import.meta.env || {}` guard. 92/92 vitest tests still pass; the production build still code-splits the locale packs the same way. **This commit alone is independently useful** — it makes everything under `frontend/src/lib/` reusable from plain Node, which the README invites ("the training logic lives in pure functions under `frontend/src/lib/`").

2. **`MCP: read-only stdio server for LLM apps to query workout history`** — Eight read-only tools, all computed by the same pure functions the React UI uses (`history.js`, `onerm.js`, `muscles.js`, `exercises.js`):

   | Tool | What it answers |
   |---|---|
   | `list_routines`, `get_routine` | The plan — names, exercise counts, per-exercise targets |
   | `get_week_plan` | What's scheduled each weekday + what applies today, accounting for override |
   | `list_workouts`, `get_workout` | Recent sessions newest-first, full per-set breakdown per date |
   | `get_bodyweight` | Weigh-ins with the goal line + signed deltas vs goal |
   | `estimate_1rm` | All-time best per exercise + trend (Epley/Brzycki/Lombardi) or a PR table |
   | `muscle_balance` | Which muscles trained vs neglected over week/month/all |

   One runtime dependency beyond the MCP SDK: none. zod is the SDK's transitive. State is read with `fs.watch` (mtime fallback for platforms where watchers are flaky) so a session the web UI just finished is visible on the next tool call. No passkey material, VAPID keys, or session secrets ever cross the boundary — handlers only serialize fields from the state object plus the user's name from `db.json`.

3. **`Comments: trim to the established cadence`** — Review pass to match the codebase's comment conventions. The MCP code initially shipped with 5–8 line file headers, JSDoc on every one-line wrapper, and inline explanations that referenced the PR itself ("so the MCP server can reuse it") — all of which `api/server.js` and `frontend/src/lib/` don't do. Trimmed across all touched files (~150 comment lines removed). Two real defects also fixed: a corrupted Cyrillic word in a `state.js` comment and a missing space in a `tools.test.js` comment.

4. **`MCP: fixes from review (cardio mode, dead code, Stryker drop, docs)`** — The follow-up after the maintainer's review:

   - **Cardio mode bug**: `entryView` was passing `e.target || { id: e.id }` to `modeOf`, which doesn't put `id` on the target when target exists — so a cardio config saved as `{sets, min, speed}` (no `mode`, no `id`) fell through to `'reps'` and the set label came back as `"0×0"`. Now uses the same pattern every app call site uses (`modeOf({ ...(e.target || {}), id: e.id })`). Same fix applied to the `setLabel` call. Regression test added using a real cardio exercise id and a target with no `mode` key — would have caught the original bug.
   - **Dead code**: `requireIso`/`ISO`/`fmt`/`friendlyDate`/`MUSCLES` were unused (zod regexes cover ISO validation at the SDK layer). Removed, along with the misplaced comment that described `requireIso` but sat above `entryView`.
   - **Forward-reference removed**: `estimate_1rm`'s `exercise_id` schema description referenced `search_exercises (v1.5)` which doesn't exist yet — an LLM reading the schema would try to call it. Dropped.
   - **resolveUid fallback**: When zero state files exist but multiple users are in `db.json`, the throw used `files.join(', ')` on an empty array and rendered `"set OPENGYM_UID to one of: "` with no ids. Now falls back to the `db.json` user ids in that branch.
   - **Stryker dropped**: The SDK's 93 transitive runtime packages are tolerable since `mcp/` is opt-in and never enters Docker, but a second test toolchain for a 39% mutation score wasn't carrying its weight next to the plain vitest suite. Removed `.stryker.conf.json`, the two Stryker devDeps, the `mutation:test` script, the `.gitignore` entries, and the README section. `mcp/` now has a single devDep (`vitest`); 0 vulnerabilities (was 2 moderate).
   - **Discoverability**: Added `mcp/` to the layout block in `CONTRIBUTING.md` and a one-line cross-reference in `README.md`'s Tech section pointing at `mcp/README.md`.

## Verification

| Suite | Result |
|---|---|
| `cd frontend && npm test` | 92 / 92 pass |
| `cd frontend && npm run build` | succeeds (locale code-splitting intact) |
| `cd mcp && npm test` | 33 / 33 pass (added one cardio regression test) |
| End-to-end stdio smoke (initialize → tools/list → tools/call × get_workout on a cardio entry with no `mode` key) | all 4 checks pass, cardio label renders as "20 min @ 8 km/h" |

## How to try it

Takes about a minute if you already use an MCP-aware LLM client:

```bash
cd mcp && npm install
```

Then add to your LLM client's config (Claude Desktop's `claude_desktop_config.json` on macOS lives at `~/Library/Application Support/Claude/claude_desktop_config.json`):

```jsonc
{
  "mcpServers": {
    "opengym": {
      "command": "node",
      "args": ["/absolute/path/to/openGym/mcp/src/index.js"],
      "env": {
        "OPENGYM_DATA": "/absolute/path/to/openGym/data",
        "OPENGYM_UID": "<your-uid>"   // optional — auto-detected on single-user instances
      }
    }
  }
}
```

Restart the client. The server logs `serving profile <name> (<uid>)` to stderr on boot, lists the 8 tools via `tools/list`, and answers JSON over stdio until the client disconnects. Find your uid in `./data/db.json` under `users[].id`.

## What this PR does *not* change

- `api/`, `web/`, `docker-compose.yml` — untouched
- Frontend behavior — untouched (92 tests still pass, production build still works)
- No new runtime dependencies in `frontend/` or `api/`
- No new container in `docker-compose.yml`
- No new auth on the api server

The only changes to existing files are the i18n split (one new file, `i18n-core.js`; two unchanged-but-import-path-changed files: `format.js` and `exercises.js` — both still import what they always did modulo the file name), a layout-block addition to `CONTRIBUTING.md`, and a one-line pointer in `README.md`.

## Out of scope (future phases, documented in `mcp/README.md`)

- **Phase 1.5:** a `progression_next` tool (what does the policy prescribe next?). No new deps; small surface area.
- **Phase 2:** read+write over stdio. Requires a long-lived token auth path minted from the admin dashboard and a write-lock against the web UI's read-modify-write of `state-<uid>.json`.
- **Phase 3:** Streamable HTTP transport as an opt-in 4th container in `docker-compose.yml`, reusing the same tool implementations behind a second transport.

## Reviewing

The four commits are independent and each passes tests on its own. Commit 1 (the i18n split) is independently useful even if you don't want the MCP server.


DuarteSantos8:
Really nice PR — reusing the actual `frontend/src/lib` helpers rather than reimplementing them is the right call, and the `i18n-core.js` split is clean with no public export changes. I ran it against my own data and the numbers match the Stats screen. Happy to merge after a few things:

1. **Cardio entries come back as `mode: "reps"`**. In `entryView` you call `modeOf(e.target || { id: e.id })`. The exercise-config sheet saves a cardio config as `{sets, min, speed}` — no `mode` key and no `id` — so `modeOf` falls through to `isCardio(undefined)` and returns `'reps'`. Everywhere else in the app passes the `id` in explicitly (`modeOf({ ...entry.target, id: entry.id })`).  
   **Repro:** add a cardio exercise mid-workout, then `get_workout` that date — you get `"mode": "reps"` and `"label": "0×0"` instead of the cardio label. Same fix applies to the `setLabel` call on the next line.

2. **Dead code in `tools.js`**: `requireIso`/`ISO` are never called (the zod regexes cover it), and `fmt`, `friendlyDate`, `MUSCLES` are imported but unused. The comment _“Drops ISO dates so an LLM passing 'yesterday' is rejected”_ is also sitting above `entryView` where it doesn't apply — I think it belongs to `requireIso`.

3. **`estimate_1rm`'s `exercise_id` description** points at `search_exercises` (v1.5), which doesn't exist yet — an LLM reading the schema will try to call it. Please drop the reference until that ships.

4. **In `state.js`**, the _“multiple openGym users found — set OPENGYM_UID to one of: …”_ throw is also reached when there are zero state files but >1 user in `db.json`, in which case the list renders empty. Worth falling back to the `db.json` user ids there.

5. **Dependencies**. The MCP SDK pulls in 93 runtime packages (`express`, `cors`, `hono`, `body-parser`…) for a server that only ever speaks stdio, plus 202 dev packages for Stryker. This project is deliberately dependency-light — `api/` has exactly one dep. I'm okay with the SDK since `mcp/` is opt-in and never enters the Docker build, but I'd like to drop the Stryker setup (`.stryker.conf.json`, the three devDeps, the two `.gitignore` entries): a second test toolchain for a 39% score isn't carrying its weight next to the plain vitest suite, which is good.

Also: please add `mcp/` to the layout block in `CONTRIBUTING.md` and a line in the root `README.md` pointing at `mcp/README.md`, otherwise nobody will find it.

The security posture is right — read-only, uid sanitised, no network, no passkey/VAPID material. Thanks for the thorough README and tests. 

Thanks a lot!!! @Pengboi

DuarteSantos8:
P. S. Sorry for the late response — I'm currently on vacation ;)

PR #21, opened 2026-07-28
Add optional RIR (reps in reserve) to reps-mode sets

octaviodimarco:
## What

Adds a third stepper column, next to weight and reps, for logging **RIR (reps in reserve)** on strength sets. Only shown for `reps`-mode sets — not cardio or timed holds, where it has no meaning.

## Why

RIR is a normal part of tracking effort for lifters running an RIR-based program (e.g. "top set to RIR 1-2"), and there's currently no way to log it alongside weight and reps.

## Design notes

- **Purely a logging/display addition.** The progression engine (linear / Greyskull LP / double progression) only ever reads/writes weight, reps and seconds — RIR never touches it, on purpose. A future PR could use it for progression decisions, but this one deliberately doesn't.
- **No migration needed.** A set only gets an `rir` key once you actually log a value — `buildSets`/`addSet` don't pre-seed it — so old workouts read exactly as before (`undefined` → not shown), matching the existing convention for optional fields in the store.
- **Round-trips automatically.** Backup export/import (`Settings.jsx`) does a whole-object `JSON.stringify`/`JSON.parse` with no field enumeration, so `rir` travels with no code changes there.
- **"Last time" summaries** now append `(RIR n)` when a set has one logged — e.g. `80×4 (RIR 1.5)` — covered by new `setLabel` tests, including the `rir: 0` case (0 is a valid RIR and must not be treated as absent, so the check is `!= null`, not falsy).
- **Mobile layout.** Fitting a third stepper next to weight+reps on narrow phones needed the same treatment as the exercise-config sheet's three-steppers-in-a-row (`.cfgrow`): narrower buttons, a tighter row gap, and a smaller value font, scoped to rows that show RIR (`.setrow.rir3`). Hand-verified at 360–390px with edge-case values (2-3 digit weight, double-digit reps, decimal RIR together) — tight but legible at the realistic worst case.
- **Auto-checks the set.** Logging RIR is naturally the last thing you do for a set — by the time you're rating reps left, you've already done it. Editing the RIR field (typing or the +/- stepper) now checks the set off itself, same beep/vibrate/rest-timer as tapping the checkbox by hand. Weight and reps don't trigger this (they're often pre-filled targets, set before the set is performed). Guarded so re-editing RIR on an already-done set doesn't re-fire the toggle side effects.

## Testing

- `npm test` — all existing tests pass unchanged, plus new cases for `setLabel` (with/without RIR, and the `rir: 0` edge case).
- Manually verified in-browser (`npm run dev`) at 360px and 375-390px viewports: layout, cardio/timed exercises correctly hide the RIR column, "Last time" summary rendering, full backup export/import round-trip, and the auto-check behavior (including that weight/reps edits don't trigger it, and re-editing RIR on a done set doesn't double-fire).

## Scope

This is step one of two — a follow-up PR will look at letting you swap an exercise for an alternative mid-workout (e.g. machine's taken). Kept separate since they're unrelated changes.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -539,6 +539,7 @@ h4.sec{
 .sethead .n-sp{width:24px;flex:none}
 .sethead .w-sp{flex:1.4;text-align:center}
 .sethead .r-sp{flex:1;text-align:center}
+.sethead .rir-sp{flex:1.1;text-align:center}
 .sethead .ck-sp{width:30px;flex:none}
 .setrow{display:flex;align-items:center;gap:8px;padding:7px 0;position:relative}
 .setrow+.setrow::before{
@@ -551,7 +552,19 @@ h4.sec{
 }
 .setrow .stp.w{flex:1.4;min-width:0}
 .setrow .stp.r{flex:1;min-width:0}
+.setrow .stp.rir{flex:1.1;min-width:0}
 .setrow .stp button{width:32px;height:40px}
+/* a 3rd stepper (RIR) needs narrower buttons and a tighter row gap to fit next to weight+reps
+   on narrow phones, same technique as the exercise-config sheet's three-steppers-in-a-row
+   (.cfgrow below) — RIR's own value is short ("1.5") so its buttons shrink the most, weight/reps
+   just enough to make room since their values ("82.5", "12") need more of what's left. */
+.setrow.rir3{gap:6px}
+.setrow.rir3 .stp.w{flex:1.2}
+.setrow.rir3 .stp.r{flex:1}
+.setrow.rir3 .stp.rir{flex:.85}
+.setrow.rir3 .stp button{width:23px}
+.setrow.rir3 .stp.rir button{width:20px}
+.setrow.rir3 .stp .num{font-size:14px}
 .setrow.done .n{background:var(--acc);color:var(--on-acc)}
 .setrow.done{opacity:.45}
 /* why the next target is what it is (issue #17) */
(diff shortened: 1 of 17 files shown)


DuarteSantos8:
Thanks — RIR is a reasonable gap to fill, and the first commit is close. The tests pass and the `rir: 0` handling in `setLabel` is the right call. Two blockers before I can take it:

- **The auto-check commit needs to go.** `NumberField.commit()` calls `onChange` on every keystroke, so typing `1.5` commits `1` first, which fires `markDoneFromRir()` → `onToggle()`. On an exercise's last set that immediately opens the top-weight sheet (or `workoutCompleteSheet` on the final unit) — a sheet pops up while you're still typing the second character, and you end up with RIR 1 instead of 1.5. Given 1.5 is your own example value, this is the common path, not an edge case.

- **Related, from the same commit:** tapping `−` on an empty RIR gives `Math.max(0, 0 - 0.5) → 0`, which auto-checks the set and logs `"RIR 0"` — a meaningful value (to failure), not `"unset"`. Clearing the field does the same, since `commit('')` yields `0`. And because `setLabel` tests `!= null`, there's no way back: once `rir` exists it can only be `0`, never `undefined`, so a mis-tap permanently stamps `(RIR 0)` on that set. I'd rather keep logging RIR and checking the set off as separate actions.

- **Please put the column behind a setting.** Right now every reps-mode row gets a third stepper, and `.setrow.rir3` shrinks the +/- buttons from 32px to 23px (20px for RIR) for everyone — including the majority who don't run an RIR-based program. A `showRir` toggle in Settings (default off) keeps the existing row untouched for them and lets the RIR layout be as tight as it needs to be for the people who opted in.

- **One more:** `t('RIR')` is a new source string with no entry in the 11 packs under `src/locales/` — please add it, even if several languages keep the acronym.

Take out the auto-check and gate the column and I'll merge it.

Thanks a lot @octaviodimarco !!!

octaviodimarco:
hey @DuarteSantos8 !

Both blockers are fixed:

- Auto-check is gone — logging RIR no longer touches the checkbox. It's back to a plain field with no side effects, same as weight/reps. 
- The column is now opt-in — a showRir toggle under Settings → Units & timer (default off). Existing rows are untouched unless someone turns it on.

Also added RIR / Show RIR column to all 11 language packs (kept RIR as the acronym everywhere, same as 1RM).

Pushed both as separate commits on top of the same branch. 

Let me know if anything else needs adjusting!

PR #38, opened 2026-08-03
Coach 1/6: server plumbing, consent, payload boundary, credential model

alexpcosta:
> **First: `coach` needs a fast-forward to `main`.** It sits at `8de55ce`, which is before
> v1.2.4, so this PR currently shows 9 commits / 52 files — 8 of them yours. `coach` is a
> strict ancestor of `main`, so `git push . origin/main:coach` is a no-conflict update, and
> this collapses to **1 commit / 25 files** the moment you do it. Branching from `8de55ce`
> instead wasn't an option: the bodyweight work is exactly what the payload had to learn.

First of six, against your PR-1 constraints. Everything here is transport and boundary — a job
runs, its answer is parsed, and **nothing yet judges whether that answer is safe to act on**.
The validator and the apply path are 2/6, and the seam is explicit: `jobs.js` carries an answer
no further than `unvalidated`, with a test pinning that so the next PR has something to replace.

## Your four constraints

**Credential ownership — per-profile, fail closed.** Three modes: `instance` (single-profile
default), `profile`, API key. In instance mode the credential binds to the first profile that
spends it and every other profile is **refused** — your message verbatim, no job enqueued, no
dismissible warning. A refused profile is never handed the token. 12 tests, including that a
stale `boundUid` left over from instance mode can't leak into per-profile mode.

Profile credentials live in `coach-auth-<uid>.json`, `0600`, deliberately **not** a field on
`state-<uid>.json` — that blob syncs and travels in the user's JSON export, and a credential in
a backup is the mistake we're avoiding, not repeating. The uid is validated against
`[A-Za-z0-9_-]{1,64}` before it reaches a path.

`docs/AI_COACH.md` carries your line: *The instance owner is responsible for their provider's
terms.*

**The `chmod 0700` on `/data`.** Gone. `secret`, `db.json` and `coach.json` carry `0600`
themselves; `state-*.json` stays readable, so #19 can read it.

One thing that turned up doing it: `db.json` was written through an `atomicWrite` with no mode,
so it was only ever protected *by* the directory seal. Removing the seal without giving it a
mode would have left passkey credential material at umask default. It now has an explicit
`0600`.

**`unprivilegedIds()` fails open.** Now `canDropPrivileges()`, and on Linux no drop means no
job — refused at enqueue. Linux-gated so a `node server.js` on a laptop still works. The admin
payload carries `unprivileged: { ok, dropped, why }` for the card.

**Codex credential path.** Not in this PR — it arrives with the Codex adapter in 6/6, outside
`./data`, not nested in it.

## Exercise catalogue

Generated copy kept, `--check` job kept and wired into CI.

On `modeOf` — it **wasn't** pinned. `coach.test.js` pins `CHANGE_TYPES`, `hashPlan` and the
payload shape, so the pattern was there, but the one function you predicted would drift was the
one with no test. `frontend/src/lib/coach-parity.test.js` now pins `modeOf`, `isBw` and
`isPerSide` behaviourally over a table of configs (169 assertions), and it earned its place
immediately — see below.

## Bodyweight and per-side

v1.2.4 landed while this was being split, and the Coach predated it. `weight` now means *added*
load on a bodyweight exercise, so it's `0` on a perfectly good session and every load-shaped
signal reads flat. Uncorrected, the Coach would have looked at a working push-up progression and
proposed adding weight to a push-up.

So: the payload carries `bodyweight`/`side`/`repsMax`, the session reader carries set count (the
dimension bodyweight work grows once reps hit the ceiling), `canonicalPlan` resolves both flags
rather than copying them so a plan that agrees with the catalogue hashes the same as one that
states it, and the prompts say reps-then-sets is the progression and per-side targets step in
twos. Your `POLICIES` list is unchanged and the validator will still be correct in 2/6.

## Dependencies

`api/` installs `@simplewebauthn/server` and `web-push`. That's all — no AI dependency until
3/6, and `@anthropic-ai/sdk`, `@modelcontextprotocol/sdk` and `zod` never arrive at all. Jobs run
against the in-repo fixture provider, so CI needs no account, no network and no image build.

## Verification

| | |
|---|---|
| `cd api && npm test` | 33 / 33 |
| `cd frontend && npm test` | 361 / 361 (169 parity) |
| `node scripts/check-locales.mjs` | 11 locales, 628 keys, in sync |
| `node scripts/build-coach-library.mjs --check` | in sync |
| `cd frontend && npm run build` | succeeds |

## Deferred, deliberately

Validator + apply + repair round (2/6) · Claude Agent SDK adapter, multi-stage Dockerfile with a
`coach` target, compose profile, `tools: []` assertion (3/6) · all UI including the
`CoachProposal` disclaimer, the account-being-spent line and the unprivileged indicator (4/6) ·
locales (5/6) · Codex (6/6).

No `openGym_AI_Strategy.pdf`, no `ai-enablement/` — the functional plan (77 `FR-xx` ids the code
references) and the design deck are linked rather than committed:
https://github.com/alexpcosta/opengym/blob/main/ai-enablement/functional-plan.md

Separately: the CI for #19's 33 tests can't be a standalone PR — `mcp/` doesn't exist on `main`
until it lands, so a job pointing at it fails on first run. I've written it up as a patch for
that PR's author instead; say the word and I'll post it on #19.

https://claude.ai/code/session_01BnXPdnec721S6ZrQJGzAfq


DuarteSantos8:
Merged. `coach` is fast-forwarded to `main` as you asked, so this landed as the single commit
it was meant to be, and your authorship is intact.

The design is right and I'm not going to re-litigate any of it: the by-name allowlist in
`payload.js` with a test that pins it against new state fields, consent enforced in `enqueue()`
rather than by the screen that collects it, the credential in its own file instead of on the
blob that syncs and gets exported, argv-not-shell, and an env built from nothing rather than
filtered. Dropping the blanket `0700` on `/data` for per-file locking is the right call for a
bind mount and I should have caught that one myself. The `unvalidated` seam with a test holding
it in place is exactly the right way to hand off to 2/6.

I pushed a fix commit on top rather than sending it back, because all three of these are the
same root cause — the api had no CI until this PR added it, so none of this code had ever been
run by anything except you.

**The CI job you added would have gone red on its own first run.** `canDropPrivileges()` fails
closed on Linux unless the process is root *and* the image has a `coach` user. A GitHub runner
is Linux, runs as an ordinary user, and has no `coach` — so every `enqueue()` there refuses with
`unprivileged`, and eight tests fail. I reproduced it under `setpriv --reuid=65534` before
touching anything. The suite now pins the verdict through a seam in `spawn.js` instead of
inheriting whatever the host would have said.

That also fixes the test that was supposed to cover this: `require()` in an ESM module, so it
threw `require is not defined` rather than asserting. And its `if (canDropPrivileges().ok)
return` meant it skipped itself on exactly the hosts where the control is load-bearing. It now
pins the refusal and asserts it.

**`testRun()` referenced a `job` that doesn't exist in its scope.** `POST /api/admin/coach/test`
threw `ReferenceError` on every call, and since the throw landed before the `try`, each click
also leaked a temp dir. No user is asking for a test run, so there's no profile that pays:
instance mode uses the bound credential, per-profile mode honestly resolves to nobody.

**And with that fixed, the round trip still failed** — the fixture answered the test prompt with
a full training plan. `payload()` finds no fence, so `kind` falls through to `create`. That
missing fence is precisely what separates the round trip from a real job, so the fixture keys on
it. So the one button that proves the runtime works had never worked with the only provider that
ships.

Two small ones while I was in there: `GET /api/coach/disclosure` was the only coach route without
a session check and told anonymous callers which provider the box is wired to — on an invite-only
instance that's not nothing. And `status()` returned no `cap` field on the expiry path only, so
the poll that happens to expire a proposal is the one where the client's cap readout goes
undefined. Push back on the disclosure gate if the consent screen needs it pre-session, but as
far as I can tell it's always behind a login.

34/34 green as root and as an unprivileged user, `build-coach-library.mjs --check` in sync at
1324 exercises, frontend 361/361 including your parity test.

Two things I left alone for 2/6, so they don't surprise you:

- There's no route to set `authMode` or to connect/clear a credential — `POST /api/admin/coach/config`
  takes `enabled`/`provider`/`model`/`caps` only. So the per-profile mode the whole credential
  model is built around can't actually be switched on yet. Fine if that's where the gap in
  `routes.js` is headed, just worth naming.
- `refine` reads `pendingCreate?.bundle`, but pending only ever carries `unvalidated`, so
  `previous` is always null. Consistent with the seam — it just starts working for free once the
  validator lands, which is easy to mistake for something that already works.

Send 2/6 whenever. Now that the api job exists it'll actually run on the PR.

alexpcosta:
Thanks — and thanks for pushing the fix on top rather than sending it back. You were right about the root cause: the api had no CI until that PR added it, so none of that code had ever been run by anything except me, and "it passes on my laptop" is exactly the guarantee `canDropPrivileges()` was designed not to give. I've adopted the `spawn.js` seam as the way the suite states its own preconditions, and 2/6 uses it the same way.

All four of your fixes are right and I'm not going to relitigate them either. Two notes, both agreeing with you:

The `require()` in an ESM module is worse than a broken test — it means the assertion had never executed anywhere, so "the control fails closed" was an untested claim in a PR whose whole argument was that it fails closed. The version you wrote, which pins the refusal and asserts it, is the test I should have written.

`testRun()` is the one path in the feature that starts without a profile, and I built every other path from one. Your reading — instance mode spends the bound credential, per-profile mode honestly resolves to nobody — is the right answer, and it is better than making the admin card pretend it can test somebody else's account.

**2/6 is ready.** It closes the `unvalidated` seam: `validate.js` with the closed list of change types, and the client-side apply/snapshot/revert path. Branched from `de7f25c`, so your commit is the parent.

Two things the seam was hiding, both mine, both only reachable once the validator landed:

`hashPlan` was still hashing the pre-1.2.4 field list. `canonicalPlan` learned `repsMax`, `bodyweight` and `side` when the payload did, and the fingerprint kept the old one — so it computed all three and discarded them. A rep ceiling raised by hand read as an untouched plan, on exactly the exercises where that ceiling is how progression works. Fixed on both sides, with a test per field.

And `refine` is worse than you framed it. You said `previous` is always null and starts working for free once the validator lands. It does — but "always null" means every refine so far asked the model to revise a plan it had never seen, and got an answer built from the intake alone. That is not a seam waiting to close, it is a feature that silently did something else. The fixture now echoes whether the previous bundle arrived, so the test can tell a real refine from a plausible one.

On your two carry-overs: the `authMode` and connect/clear routes go in 3/6, not 2/6 — they exist to serve a credential and the only provider this build ships is the fixture, which has none, so `credentialFor` returns before it ever consults the mode. A switch whose two positions do the same thing seemed worse than no switch. There's a comment at the gap in `routes.js` naming where it went.

No pushback on the disclosure gate. I checked rather than assumed: every Coach entry point sits behind `coachAvailable(config, user, …)`, which requires a user outside the demo build, and the consent screen is inside that gate — `Coach.jsx` redirects before `ConsentCard` mounts. In the demo build `disclosure()` short-circuits to a local fixture and never reaches the server. Your gate is safe on both paths.

One call I'd like you to look at, since I could see it going the other way: `swap-exercise` now drops an explicit `bodyweight` or `side` flag, on the grounds that those describe the movement rather than the prescription. `bodyweight` then falls back to the catalogue, which is right. `side` has no catalogue fallback, so swapping one unilateral exercise for another loses the flag and the user re-sets it. I took that over the alternative — carrying `side: true` onto a bilateral movement makes the app halve a rep count that was never per-side — but it is a judgement call, not a fact, and it is one line to reverse.

65/65 on the api, run on macOS and on Linux both as root and as an unprivileged user with no `coach` user in the image; 403/403 on the frontend; locales in sync; library in sync at 1324. Now that the api job exists, it should go green on the PR on its own.

PR #46, opened 2026-08-06
Coach 3/6: the Claude provider, and an AI runtime you only carry if you ask

alexpcosta:
**3/6 — the Claude Agent SDK adapter, its dependency, and the opt-in packaging.** Stacked on [#45](https://github.com/DuarteSantos8/openGym/pull/45); it collapses to a single commit once that merges.

## Two build targets

`api/Dockerfile` now builds two images from one file:

| target | what it carries | size |
|---|---|---|
| `default` (no `--target`) | no AI runtime, no AI dependency — `npm ci --omit=optional` | ~158 MB |
| `coach` | the same image plus `@anthropic-ai/claude-agent-sdk` | ~455 MB |

```
docker build --target coach ./api
API_TARGET=coach docker compose up --build api
```

The SDK's runtime ships as a **per-platform package**, so on this Alpine base the lockfile resolves `linux-x64-musl` / `linux-arm64-musl`. That is exactly the class of thing that builds fine and fails on somebody else's machine — so the new `api-image` job builds **and runs** both targets: the default has to boot and serve `/api/config` with the SDK genuinely unimportable, and the coach one has to report a runnable SDK through the same `check()` the admin card renders.

## The `coach` user was missing, and it disabled everything

Both targets now `adduser coach`. This is not a Claude detail. `canDropPrivileges()` fails closed, so in an image without that user **every** job is refused with `unprivileged` — including the fixture provider that exists precisely so the loop can be walked before an account is connected. Between this and #45, the shipped api image could not start, and would not have run a job if it had.

The `api-image` job asserts `id coach` for the same reason it asserts the boot: both are things the source tree cannot tell you.

## What the model is allowed to do

Four SDK options are why this adapter can sit next to someone's `./data`:

- `tools: []` — the SDK's documented *"disable all built-in tools"*. No Read, no Bash, no Grep; there is no filesystem tool to talk it into using.
- `settingSources: []` — no user/project/local settings file is loaded, so nothing on the host widens that afterwards.
- `skills: []` — same, for skills.
- `strictMcpConfig` — no MCP server from ambient config.

They are one frozen exported object asserted **by value** in `adapters.test.js`, so re-enabling one is a red build rather than a quiet capability grant. This is the test you asked for in Q3 of the original thread.

The credential reaches the model only as an environment variable: `jobEnv()` builds the environment from nothing, and the SDK's `env` option *replaces* the child environment rather than extending it — so `RP_ID`, `ADMIN_UIDS` and the VAPID keys cannot arrive by inheritance.

## Verified against the SDK's types, not from memory

I installed `@anthropic-ai/claude-agent-sdk@0.3.223` and read its type definitions before writing the adapter. Two things the earlier draft had wrong:

- **`SDKResultError` has no `errors` field.** `message.errors?.join('\n')` was dead code that always fell through to its own fallback string. The failure reason now comes from `subtype` (`error_during_execution`, `error_max_turns`, …) and `stop_reason`.
- **The client-app version string was hardcoded to `1.2.3`**, two releases after that stopped being true. It is read from `package.json` now — a version typed by hand is a version that goes stale.

`tools: []`, `permissionMode: 'dontAsk'`, `settingSources`, `skills`, `persistSession` and `strictMcpConfig` all checked out as used.

## An absent runtime is a state, not a crash

The adapter imports the SDK lazily. On the default image the module still loads, `check()` reports the runtime missing, `isConnected()` goes false, and `/api/config` carries no `coach` key — so no Coach UI is mounted anywhere. That is what makes "the default image is unchanged for people who never wanted this" true rather than aspirational.

## The routes I owed you

`POST /api/admin/coach/connect` and `/disconnect` land here, as I said in 2/6 they would: they exist to serve a credential, and until this PR no provider had one. The token is accepted once, encrypted, and leaves only as an environment variable on a job's child process; `type` must match a variable the configured provider actually declares, so a key cannot be filed under the wrong provider and silently go nowhere at job time.

Still deliberately absent: `authMode` and the per-profile connect/clear pair. Instance mode is the whole of what these two routes serve, and per-profile needs its own pair against `saveProfileAuth`/`clearProfileAuth` — that lands with the screen that drives it.

## Verified

- api **72/72** on macOS, and on Linux as uid 65534 with the SDK absent (the default-image shape).
- Both image targets built and **run**: default boots and serves `/api/config` with the SDK unimportable; coach reports `Claude Agent SDK 0.3.223`.
- `build-coach-library.mjs --check` in sync at 1324.

## Not in this PR

Codex is 6/6, so no `codex.js` here and no `bubblewrap` — it is a Codex requirement and the tree stays free of it until then. The UI is 4/6; locales 5/6.

One thing I did **not** take, in case you want it: the SDK exposes `maxBudgetUsd`, which would cap a job in dollars rather than in job count like the existing caps. It is a different axis from `perProfileDaily`/`instanceDaily` and I did not want to invent policy — say the word and it is a small follow-up.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,3 +61,45 @@ jobs:
       # that forgets to regenerate it would otherwise only surface as "the Coach can't find
       # that exercise" on somebody's instance.
       - run: node scripts/build-coach-library.mjs --check
+
+  # Builds *and runs* both image targets. Building alone would not have caught the api image
+  # shipping without coach/ — that image built perfectly and exited on boot. So each target has
+  # to answer for itself: the default one starts and serves, and the coach one additionally
+  # resolves the Agent SDK's per-platform runtime (musl on this base), which is the one thing
+  # that cannot be checked from the source tree.
+  api-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - run: docker build --target default -t opengym-api:default ./api
+      - run: docker build --target coach   -t opengym-api:coach   ./api
+
+      - name: the default image boots and serves, with no AI runtime in it
+        run: |
+          docker run -d --name api-default -e DATA_DIR=/data -p 3000:3000 opengym-api:default
+          for i in $(seq 1 30); do
+            curl -fsS http://localhost:3000/api/config >/dev/null 2>&1 && break
+            sleep 1
+          done
+          curl -fsS http://localhost:3000/api/config | tee /dev/stderr | grep -q '"invite_only"'
+          # Absent by construction, not by accident: --omit=optional is the only thing keeping
+          # the AI dependency out of the image an owner gets when they never asked for it.
+          docker run --rm opengym-api:default node -e \
+            'import("@anthropic-ai/claude-agent-sdk").then(()=>{console.error("AI runtime present in the default image");process.exit(1)},()=>process.exit(0))'
+          docker rm -f api-default
+
+      - name: the coach image carries a runnable Agent SDK
+        run: |
+          docker run --rm opengym-api:coach node --input-type=module -e '
+            const { adapterFor } = await import("./coach/adapters/index.js");
+            const r = await adapterFor("claude").check();
+            if (!r.ok) { console.error(r.error); process.exit(1); }
+            console.log(r.version);
+          '
+
+      - name: the privilege drop can actually be performed in the image
+        run: |
+          # canDropPrivileges() fails closed, so a missing `coach` user disables every job —
+          # including the fixture provider. Cheaper to assert here than to discover on an instance.
+          docker run --rm opengym-api:default id coach
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /app
 COPY package.json package-lock.json* ./
 RUN npm install --omit=dev && npm cache clean --force
 COPY server.js ./
+COPY coach ./coach
 ENV NODE_ENV=production
 EXPOSE 3000
 CMD ["node", "server.js"]
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -1,9 +1,47 @@
-FROM node:22-alpine
+# Two targets from one file.
+#
+#   default (built when no --target is given) — what every instance has always got. No AI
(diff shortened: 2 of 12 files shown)


DuarteSantos8:
Merged, thanks. This one I checked properly rather than taking the description on trust, because
the whole thing rests on four option names being real.

I pulled `@anthropic-ai/claude-agent-sdk@0.3.223` and read its type definitions. All four are
genuine `Options` fields and the semantics are exactly as you describe them — `tools: []` is
documented verbatim as *"`[]` (empty array) - Disable all built-in tools"*, `settingSources: []`
as *"Pass `[]` to disable filesystem settings (SDK isolation mode)"*, and `env` as *"REPLACES the
subprocess environment entirely — it is not merged with `process.env`"*. That last one is why
`jobEnv()` supplying its own `PATH`/`HOME`/`TMPDIR` matters, and you did that.

Your correction was also right: `SDKResultError` has `subtype` and `stop_reason` and no `errors`
field. Asserting `LOCKDOWN` by value in `adapters.test.js` is the right shape for this — a
capability grant should be a red build, not a quiet diff.

Verified at the merge commit: api 72/72, frontend 404/404, clean build, 11 locales in sync, and
both image targets built and run — default boots with the SDK genuinely unimportable, `coach`
target boots with a runnable SDK, and `id coach` resolves in both (uid 1001). Catching that the
missing `coach` user would have failed *every* job closed, not just the Claude one, was a good
find.

One correction for the description, since it'll get quoted later: the size table doesn't hold as
written. On my box (containerd snapshotter) the default image is **243 MB** on disk / 58.8 MB
compressed, and the coach one **677 MB** / 154 MB — neither is the claimed 158 MB / 455 MB. The
default is essentially `node:22-alpine` (230 MB) plus ~13 MB of app and prod deps, so 158 MB
isn't reachable on any measure here. What *is* accurate is the delta you were actually making a
point about: the coach image's unique, non-shared size is **434.5 MB** against the default's
12 kB — so "the SDK plus its musl runtime costs ~434 MB, and you only carry it if you ask" is
exactly right. Probably a compressed-vs-uncompressed mix-up, or measured on a classic daemon.
The mechanism is sound either way — `optionalDependencies` + `--omit=optional` does drop it
cleanly, which I confirmed by the import probe rather than by reading the lockfile.

PR #47, opened 2026-08-06
Coach 4/6: the screens, and the four things they were supposed to say

alexpcosta:
**4/6 — the UI.** Stacked on [#46](https://github.com/DuarteSantos8/openGym/pull/46) (and so on [#45](https://github.com/DuarteSantos8/openGym/pull/45)); it collapses to a single commit as those merge.

Coach hub, six-screen intake, proposal review with per-change accept/reject, and the admin card — wired into `App.jsx` and `Admin.jsx`. `AdminCoach` renders nothing unless the instance offers the Coach, so an admin page on a box that never enabled it is byte-for-byte the page it was.

## The four things I owed you

**The `CoachProposal` disclaimer.** I flagged this in the first round: the hub and the intake screen both carry a health note and the *accept* screen did not — the one place it matters, because that tap changes what somebody does with their body next week. It now sits immediately above the accept button on both proposal kinds, not at the top of a scrolling page, so it is on screen at the moment of the decision.

**"Jobs run unprivileged: yes/no."** The card never showed it. `canDropPrivileges()` fails closed, so when the answer is no *nothing runs at all* — and the admin was left to learn that from a user reporting that the Coach does nothing. It now states the verdict, and the server's own `why` when it is no.

**Whose account is being spent.** Your constraint was that the admin card and the user's Coach screen must both name it. That was in neither. `coach-api.js` gained the `/api/coach/account` call — the route 1/6 built and nothing had used since — and the card states the mode plainly, including that in instance mode every profile after the first is refused.

**The refusal path** is reachable from the client now rather than being a server-side string nobody could see.

## Something worse than a gap

The credential UI was wired to a server shape that does not exist. It read `d.auth.state`, which `GET /api/admin/coach` never sent, and posted to `/auth/setup-token`, `/auth/key` and `/auth/disconnect` — none of which are routes. **Every credential button on the card was dead.** It now uses the `connect`/`disconnect` routes from 3/6, and the admin payload carries an `auth` summary: state, type, and the account label, never the credential itself. `unreadable` is its own state rather than folding into "not connected", because `./data` restored without its `secret` has a specific cause and a specific fix.

The setup-token sheet also gained an optional "whose account is this?" field — that label is what both surfaces show when they name the account.

## Left out on purpose

The ChatGPT device-login sheet. It is Codex's, it calls routes that do not exist, and Codex is 6/6 — so it comes back with the provider it belongs to rather than shipping here as unreachable UI.

## Verified

api **72/72**, frontend **404/404**, build clean, 11 locales in sync at 628 keys. The new user-facing strings fall back to English until 5/6 adds them — the locale gate compares the eleven files against each other, so it passes either way.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,3 +61,45 @@ jobs:
       # that forgets to regenerate it would otherwise only surface as "the Coach can't find
       # that exercise" on somebody's instance.
       - run: node scripts/build-coach-library.mjs --check
+
+  # Builds *and runs* both image targets. Building alone would not have caught the api image
+  # shipping without coach/ — that image built perfectly and exited on boot. So each target has
+  # to answer for itself: the default one starts and serves, and the coach one additionally
+  # resolves the Agent SDK's per-platform runtime (musl on this base), which is the one thing
+  # that cannot be checked from the source tree.
+  api-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - run: docker build --target default -t opengym-api:default ./api
+      - run: docker build --target coach   -t opengym-api:coach   ./api
+
+      - name: the default image boots and serves, with no AI runtime in it
+        run: |
+          docker run -d --name api-default -e DATA_DIR=/data -p 3000:3000 opengym-api:default
+          for i in $(seq 1 30); do
+            curl -fsS http://localhost:3000/api/config >/dev/null 2>&1 && break
+            sleep 1
+          done
+          curl -fsS http://localhost:3000/api/config | tee /dev/stderr | grep -q '"invite_only"'
+          # Absent by construction, not by accident: --omit=optional is the only thing keeping
+          # the AI dependency out of the image an owner gets when they never asked for it.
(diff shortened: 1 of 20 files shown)


DuarteSantos8:
Merged, thanks — and thank you for going back for the four things rather than waiting to be
asked twice.

The disclaimer placement is the one I care about most: above the accept button on both proposal
kinds, not at the top of a scrolling page. That is the moment it means anything.

Finding that the credential UI was wired to a server shape that never existed — reading
`d.auth.state` and posting to three routes that aren't routes — is the kind of thing that only
turns up when someone actually tries to use it. Good catch, and `unreadable` deserving its own
state rather than folding into "not connected" is right: it has a specific cause and a specific
fix.

I checked the server side of this one closely since it touches the admin payload: the new `auth`
summary returns state, type and label only, and the `decrypt()` call is used purely as a
readability probe with its result discarded, so no credential can reach the client. Good.

Verified at the merge commit: api 72/72, frontend 404/404, clean build, 11 locales in sync.
Leaving the ChatGPT device-login sheet for 6/6 was the right call.

`coach` is still an integration branch — nothing here reaches users until 5/6 and 6/6 land.

Issue #33, opened 2026-08-01
Progression for bodyweight exercises


Alephas:
### What problem does this solve?

I would like to plan progression with sets/reps, not weight.

### Proposed approach

Configure progression by set/rep, not weight.

### Dependencies

- [ ] This doesn't require adding a new dependency

DuarteSantos8:
Rolling this into one change with #32 and #31 — details there: [[My comment](https://github.com/DuarteSantos8/openGym/issues/31#issuecomment-5151811892)]

Short version: bodyweight rep-progression already exists in `progression.js`, it just triggers on a logged weight of 0 rather than on the exercise being marked bodyweight. What's missing from your request is the sets dimension and a proper rep ceiling, which I'm adding alongside.

DuarteSantos8:
#33 — progression for bodyweight exercises

Shipped in v1.2.4.

Partly this already existed and you couldn't reach it: progression skipped load and added a rep whenever the logged weight was zero, but it keyed off what you logged, not off the exercise being bodyweight — so you'd have had to type a 0 to get there. Fixing #32 opened that path, which is why the two shipped together.

What's actually new is the ceiling. Set a top of the range and reaching it puts the reps back to the bottom and adds a set instead of climbing forever; at six sets it stops and says what it should have said earlier, which is that it's time for added weight or a harder variation. Leave the ceiling unset and it behaves exactly as before, one more rep per clean session.

PR #48, opened 2026-08-06
Session UX: warm-up phase separation, warm-up set management, weight cascade

Space-Hermes:
Three small session-screen improvements that came out of daily use:

- **Warm-up / work phase separation** - warm-up sets now render under their own "Warm-up" header with a divider before the work sets, so a session reads clearly in two phases.
- **Add / remove warm-up sets directly** - a new "Add warm-up set" button inserts before the first work set, and each warm-up row has its own remove button. The warmup flag is stored on the set and survives into history.
- **Weight cascade** - editing a set weight carries it forward to the following sets of the same phase, which saves retyping every row when you bump the bar.

All three are suggestions - happy to adjust wording, behaviour or scope. No linked issue exists for these (checked the open ones). Tests: 192/192 vitest, production build green, base 551b072c.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -559,6 +559,9 @@ h4.sec{
 .setrow+.setrow::before{
   content:'';position:absolute;left:32px;right:0;top:0;height:var(--hair);background:var(--sep);
 }
+.setph{margin:8px 0 2px;font-size:11px;letter-spacing:.08em;text-transform:uppercase;color:var(--dim);display:flex;align-items:center;gap:10px}
+.setph::after{content:'';flex:1;height:1px;background:var(--sep)}
+.setsep{height:1px;background:var(--sep);margin:9px 0}
 .setrow .n{
   width:24px;height:24px;border-radius:50%;background:var(--surface-2);color:var(--label-2);
   display:flex;align-items:center;justify-content:center;font-size:12px;font-weight:500;flex:none;
--- a/frontend/src/lib/effort.js
+++ b/frontend/src/lib/effort.js
@@ -44,7 +44,7 @@ export const scaleName = kind => EFFORT[kind].hd
 function eachDoneSet(S, fn) {
   ;(S.workouts || []).forEach(w =>
     (w.entries || []).forEach(e =>
-      (e.sets || []).forEach(s => { if (s.done) fn(s, w, e) })))
+      (e.sets || []).forEach(s => { if (s.done && !s.warmup) fn(s, w, e) })))
 }
 
 // A window in days, counted back from now. 0 = everything, which is also what an empty
--- a/frontend/src/lib/history.js
+++ b/frontend/src/lib/history.js
@@ -270,3 +270,49 @@ export function streakWeeks(S) {
   }
   return streak
 }
+
+/**
+ * Cascade a weight change forward: following sets of the same warm-up flag that are still
+ * undone take the new value (null deletes the key). Done sets are never rewritten.
+ */
+export function cascadeWeight(rows, from, value) {
+  const warm = !!rows[from]?.warmup
+  const next = rows.slice()
+  for (let j = from + 1; j < next.length; j++) {
+    if (!!next[j].warmup === warm && !next[j].done) {
+      if (value == null) delete next[j].w
+      else next[j].w = value
+    }
+  }
+  return next
+}
+
+/** Insert a warm-up row before the first work row, copying the preceding warm-up's values. */
+export function insertWarmupRow(rows, mode, target) {
+  const firstWork = rows.findIndex(x => !x.warmup)
+  const at = firstWork === -1 ? rows.length : firstWork
+  const l = rows[at - 1] || rows[rows.length - 1]
+  const warm = mode === 'cardio'
+    ? { min: l ? l.min : (target.min || 20), speed: l ? l.speed : (target.speed || 8), done: false, warmup: true }
+    : mode === 'time'
+      ? { sec: l ? l.sec : (target.sec || 45), w: l ? (l.w || 0) : (target.weight || 0), done: false, warmup: true }
+      : { w: l ? l.w : 0, r: l ? l.r : target.reps, done: false, warmup: true }
+  const next = rows.slice()
+  next.splice(at, 0, warm)
+  return next
+}
+
+/** Remove the row at `i`, never emptying the entry below one row. */
+export function removeRowAt(rows, i) {
+  if (rows.length <= 1) return rows.slice()
+  const next = rows.slice()
+  next.splice(i, 1)
+  return next
+}
+
+/** Completed non-warm-up sets across a workout's entries. */
+export function workSetsDone(w) {
+  return (w?.entries || []).reduce(
+    (n, e) => n + (e.sets || []).filter(s => s.done && !s.warmup).length, 0,
+  )
+}
--- a/frontend/src/lib/history.test.js
+++ b/frontend/src/lib/history.test.js
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest'
-import { modeOf, isTimed, fmtSec, setLabel, defaultConfig, buildSets, freestyleConfig, exLine, workoutVolume, effortOf, stepEffort, capEffort, isBw, isPerSide, sideReps, repStep } from './history.js'
+import { modeOf, isTimed, fmtSec, setLabel, defaultConfig, buildSets, freestyleConfig, exLine, workoutVolume, effortOf, stepEffort, capEffort, isBw, isPerSide, sideReps, repStep, cascadeWeight, insertWarmupRow, removeRowAt } from './history.js'
 import { EXDB } from './exercises.js'
 
 // Real ids out of the shipped catalogue, so the body-part fallback is exercised for real.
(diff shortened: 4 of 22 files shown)


DuarteSantos8:
Thanks — all three of these come from real use and it shows; the warm-up header with a divider
reads well. It's on **https://gym-test.duarte-santos.ch** to try.

Two things need fixing first.

**1. The `warmup` flag is stored and persisted, but nothing downstream reads it.** Every consumer
still filters on `s.done` alone, so a warm-up set you tick off counts as real training
everywhere:

- `lib/history.js:214` — training volume
- `lib/history.js:219` / `:224` — set counts
- `lib/effort.js:47` — effort statistics
- `lib/muscles.js:119` — muscle-map load
- `lib/onerm.js:47` — 1RM estimation
- `lib/progression.js:113` / `:122` / `:253` — next session's prescription

The description calls "survives into history" a feature, but until those read the flag it means
warm-ups inflate your volume, your muscle map and what the app tells you to lift next week.
Filtering `!s.warmup` alongside `s.done` in those places is most of the fix.

**2. The weight cascade overwrites sets that are already done.** `setField` walks forward
unconditionally, so correcting set 1's weight after you've finished the exercise rewrites sets
2–4 — changing the record of what you actually lifted, with no undo. Skipping `done` sets when
cascading would keep the convenience without the data loss. Worth considering pyramid/drop-set
users too, who deliberately want different weights per row.

Also: this adds no tests (192/192 is the pre-existing count), and warm-up insertion, removal and
the cascade are all logic worth pinning down.

Note it now conflicts with #50, which has since merged — both touch `Workout.jsx`. It rebases
cleanly, no overlapping hunks.

DuarteSantos8:
Thanks — the cascade fix is right, and I have kept this on
**https://gym-test.duarte-santos.ch** so it stays testable. Not merging yet, because the warm-up
half is a deliberate half.

**The cascade is settled.** Skipping `done` sets means correcting an early row no longer rewrites
what you actually lifted. That was the data-loss concern and it is gone.

**The warm-up flag is still read by only half its consumers.** Your commit says it plainly —
*"excluded from 1RM and progression only (volume semantics kept)"* — so this is a decision rather
than an oversight, and it is a defensible one for tonnage. But it is applied wider than tonnage:

- `history.js:214` — volume. Defensible: you did lift it.
- `history.js:219` / `:224` — set counts. A headline "sets done" that includes warm-ups is at
  least arguable.
- `effort.js:47` — **this is the one I would not keep.** The Stats card reads "X of Y finished
  sets rated". Warm-up sets are the sets nobody rates, so they only ever grow Y. Adding warm-ups
  makes your own rating coverage look worse the more diligently you log them.
- `muscles.js:119` — muscle-map load. A warm-up set adds exactly as much strain as a top set.

If "volume semantics kept" is the intent, I would still carve out the effort denominator and the
muscle map, because neither is a volume measure — one is a coverage ratio and the other is strain.

**This now interacts with #44, which merged today.** The new Fatigue view derives its stimulus
from completed sets (`recovery.js:72` takes any `done === true`, with no warm-up check). So with
this PR in, adding warm-up sets inflates the fatigue map too — a muscle reads more fatigued the
more thoroughly you warm it up, which inverts what the view is for. Neither PR is wrong on its
own; together they are. Worth deciding deliberately rather than discovering later.

**One line is dead code.** `Workout.jsx:190`:

```js
if (cfg.weightPrescription?.kind === 'percentage') return
```

`weightPrescription` does not exist anywhere in this project — not on `main`, not elsewhere in
this branch. The optional chaining means it never throws, it just never fires either, so the guard
protects against a feature that isn't there. Either it belongs to something not in this PR, or it
should come out.

**Still untested:** only `onerm.test.js` gained cases. Warm-up insertion, warm-up removal and the
cascade — including the new "never overwrite a done set" rule, which is the part that protects
data — have no coverage. That rule especially is worth pinning down now that it exists.

Space-Hermes:
Thanks — the effort-card point settles the carve-outs. Agreed on all three:

- **effort denominator**: warm-ups excluded — the ratio reflects rated working sets only.
- **muscle map** (balance "by sets worked" + strength strain): excluded — warm-ups are preparation, not effective stimulus (Dankel et al. 2017: only the near-failure reps drive the adaptation view).
- **1RM / progression**: already excluded.

On the fatigue interaction with #44 — settled deliberately, and it splits on the science: warm-up sets are real mechanical work (Bishop 2003; the impulse-response literature), so excluding them from the fatigue stimulus would under-predict recovery for high-ramp sessions. What I found is that the current intensity weight overweights light sets ~2x (the inline e1RM is itself low) — that is what would have inflated the Fatigue view. I'm re-basing the intensity on the lifter's own best weight for the exercise, with the inline estimate as the fallback until there is history. The equation is the same for everyone — no sport-specific logic — it just responds to the actual loads: four light warm-ups ≈ one top-set-equivalent, while a long ramp (e.g. ten sets at 55–75%, common in Olympic lifting but equally valid for anyone) ≈ 2.5–4x. The view reads honest for every style.

**Set counts**: showing both — total sets and work sets — in the session summary and the volume table, so nothing is hidden.

Dead guard removed (the `weightPrescription` check); tests added for warm-up insertion/removal and the cascade's never-overwrite-a-done-set rule.

DuarteSantos8:
Thanks — the carve-outs are right and the helper tests are what was missing. Two of the new exclusions are wrong though, and both are the same failure mode as last round: a filter applied at some read points but not others, and logged data getting rewritten.

**1. `applyPrescription` now overwrites *done* warm-up sets.** `progression.js:253`:

```js
const out = sets.map(s => {
  if (s.done && !s.warmup) return s     // was: if (s.done) return s
```

Its own docstring, three lines above, says "only on sets that have not been logged yet". A warm-up you have already ticked off now falls through and gets rewritten with the working prescription's weight and reps — the data-loss concern that got the cascade fixed, reappearing two files over. The condition is inverted: warm-ups should be skipped regardless of `done`, so `if (s.done || s.warmup) return s`.

**2. `readSession` filters warm-ups out of `weight` and nothing else.** `:110` and `:119` gained the guard, but in the same return object:

- `held` / `reps` are built from `sets.map(...)` over every row, so a warm-up contributes its own duration/reps, or a `0` when undone.
- `ok = ... held.every(h => h >= goal)` / `reps.every(r => r >= goal)` — **one warm-up row makes `ok` false permanently, so progression never advances again.** That is the headline one.
- `low = Math.min(...reps)` — a warm-up's lighter reps drag the double-progression floor down.
- `count = reps.length` — inflated, and `count` is the dimension bodyweight progression grows (#33).

`applyPrescription` has the same shape: `if (p.sets > out.length)` now compares against a length including warm-up rows, so adding warm-ups suppresses the bodyweight set growth. Filtering once at the top — `const sets = ((entry && entry.sets) || []).filter(s => !s.warmup)` — is cleaner than guarding each read.

**3. Needs a rebase.** This now conflicts with `main` (`index.css`, `history.js`, `history.test.js`, `muscles.js`, the locale files) after the #44/#49/#51 merges and the v1.2.6 changelog. GitHub also reports no check runs at all on the current head `1354096`, so the 192/192 isn't verified upstream.

**Nits.** Set numbering runs straight through the divider, so with two warm-ups your first work set reads "3" — worth deciding deliberately now that the phases are visually split. The `{0} sets · {1} work` string is fine and all 11 locales have it.

On the #55 interaction: its head commit already excludes warm-ups from `strengthOf` while keeping their load-weighted volume in the fatigue stimulus, which settles that thread. The two PRs touch different hunks of `muscles.js`, so whichever lands second just needs the rebase.

Space-Hermes:
Thanks — all three points are addressed on the rebased branch (head now `01072ef`).

**1. `applyPrescription` no longer touches warm-up rows.** The guard is now `if (s.done || s.warmup) return s` — a logged set is never rewritten, and neither is a warm-up in any state. The set-growth path got the same treatment: `p.sets` now compares against the work rows only (warm-ups no longer suppress the bodyweight set growth), and new rows are seeded from the last work row.

**2. `readSession` filters once at the top.** `const sets = ((entry && entry.sets) || []).filter(s => !s.warmup)` — so `held`/`reps`, `ok`, `low` and `count` all read work rows only. The headline case is pinned by a test: an undone warm-up no longer makes `ok` false permanently, and a warm-up's lighter reps no longer drag the double-progression floor. Same treatment for the time-mode `held` path.

**3. Rebased onto current main** (v1.2.6 — includes the #44/#49/#50/#51 merges). The only conflicts were the import lines in `Workout.jsx` and `history.test.js`, resolved by keeping both sides (#50's `freestyleConfig` alongside the warm-up helpers). Full suite is **237/237** and the production build is green on the new head. I left the old branch commit in place rather than anything destructive — the PR head now points at the rebased history.

**Numbering** — decided deliberately with the visual split: set numbers now restart per phase, so with two warm-ups your first work set reads 1, not 3.

New tests cover: warm-up exclusion in `readSession` (reps + time), `applyPrescription` leaving a done warm-up exactly as logged, and the work-row-only growth. Happy to adjust anything else.

DuarteSantos8:
Thanks — checked the rebased head `01072ef` against the code rather than the summary, and all three land.

**`applyPrescription`** — `if (s.done || s.warmup) return s` is right, and the test pinning a done warm-up as byte-identical (`{ w: 20, r: 8, done: true, warmup: true }`) is the one I wanted to see. The growth path counting work rows only is the fix for the #33 interaction I didn't spell out.

**`readSession`** — filtering once at the top is the right shape, and I like that you took the now-redundant `!s.warmup` guards back off `weight` rather than leaving them in as belt-and-braces; a reader can see at a glance that `sets` means work rows. `held`/`reps`/`ok`/`low`/`count`/`amrap` all read the same thing now, and the time-mode test asserting `ok === false` because the 30s *work* row missed — not because a warm-up was present — pins exactly the distinction that was broken.

**The rebase** is clean and #50's `freestyleConfig`/`preferLast` path came through intact. 237/237 and a green build reproduce on my side.

Per-phase numbering was the right call over renumbering continuously.

**One thing left, and it is not a blocker.** The growth loop terminates only if the seed is not a warm-up row:

```js
const seed = workRows[workRows.length - 1] || out[out.length - 1]
while (out.filter(s => !s.warmup).length < p.sets) out.push({ ...seed, done: false })
```

With `workRows` empty the seed is a warm-up, `{ ...seed }` carries `warmup: true`, the filtered length never grows, and it spins forever. It is unreachable today — both callers pass `buildSets()` output and `buildSets` builds fresh `{ w, r, done }` objects with no `warmup` key — but an all-warm-up entry *is* reachable in the UI: add a warm-up to a one-set exercise, then the generic "Remove set" pops the last row and leaves `[warm]`. Now that `warmup` is persisted and flows through history, that gap is worth closing before something downstream passes stored sets in. Either strip the flag on the seed, or guard the branch on `workRows.length`.

Merging this. Thanks for the three rounds — the warm-up flag reads consistently everywhere now, which is what it needed.

PR #49, opened 2026-08-06
Rest timer phone notification (local, no new deps)

Space-Hermes:
When the rest timer finishes while you are in another tab or app, openGym now shows a system notification ("Rest over — next set!") on top of the usual beep. Permission is requested lazily the first time a rest starts, and everything degrades gracefully where notifications are unsupported or denied.

Implementation notes:
- Local Web Notifications API only - no new dependencies, matching the project’s dependency-light stance.
- The existing server push path (web-push) is untouched.
- The Capacitor mobile shells would need a native plugin for true background notifications - a possible follow-up, out of scope here.

A suggestion - happy to adjust wording, behaviour or scope. Tests: 192/192 vitest, production build green, base 551b072c.

### DIFF
--- a/frontend/src/store/useUI.js
+++ b/frontend/src/store/useUI.js
@@ -10,6 +10,35 @@ import { useStore } from './useStore.js'
 const pushRestTimer = sec => { if (useStore.getState().user) api('/api/push/rest-timer', { method: 'POST', body: JSON.stringify({ seconds: sec }) }).catch(() => {}) }
 const cancelPushRestTimer = () => { if (useStore.getState().user) api('/api/push/rest-timer/cancel', { method: 'POST', body: '{}' }).catch(() => {}) }
 
+const notificationsSupported = () => typeof window !== 'undefined' && 'Notification' in window
+let requestRestNotificationPermissionP = null
+
+const requestRestNotificationPermission = async () => {
+  if (!notificationsSupported()) return false
+  if (Notification.permission === 'granted') return true
+  if (Notification.permission === 'denied') return false
+  if (!requestRestNotificationPermissionP) {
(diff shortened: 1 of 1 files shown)


DuarteSantos8:
Thanks — lazy permission request, graceful degradation, no new dependencies, and leaving the
existing web-push path alone is all the right instinct. On **https://gym-test.duarte-santos.ch**
to try.

Two gaps:

**1. It won't work on the platform in the title.** `new Notification()` throws
`Illegal constructor` on Android Chrome — that platform requires
`ServiceWorkerRegistration.showNotification()`. Your `try/catch` swallows it, so on the most
common phone case the feature silently does nothing. There's already a service worker (`sw.js`)
to hang this off.

**2. `t('Rest over')` is a new string that exists in zero locale files.** The locale gate only
compares the eleven locales against each other, so it passes CI, but the notification title ships
untranslated in all 11 languages. `'Rest over — next set!'` is fine — that one already exists for
the toast.

Both are small. Worth doing, because the feature is genuinely useful when the tab is backgrounded.

DuarteSantos8:
Merged — both points addressed, thanks.

The service-worker path is the right shape: `getRegistration()` first, direct constructor as the
fallback. The registration exists (`main.jsx:13`), so it will actually be taken on the phones that
need it rather than silently falling through.

Reusing the existing `'Rest over — next set!'` string instead of adding a new key is a better
answer than the one I asked for — it removes the translation gap rather than filling it, and no
locale file had to change at all.

One cosmetic thing for whenever you are next in there: the title and the body are now the same
sentence, so the notification prints it twice. Dropping the `body` entirely, or pairing it with a
different existing string, would read better on the lock screen.

Live on https://gym.duarte-santos.ch in v1.2.6.

PR #50, opened 2026-08-06
Freestyle sessions prefill from your last workout

Space-Hermes:
Small quality-of-life change: when you add an exercise to a freestyle (empty) workout, its sets are pre-filled from your last logged session of that exercise - same sets, reps and weight (or duration/speed for cardio, seconds for holds) - so the session starts from where you left off instead of from blanks.

Routine-driven sessions keep the existing progression logic untouched.

A suggestion - happy to adjust wording, behaviour or scope. Tests: 196/196 vitest (4 new for the prefill helper), production build green, base 551b072c.

### DIFF
--- a/frontend/src/lib/history.js
+++ b/frontend/src/lib/history.js
@@ -152,6 +152,21 @@ export function lastEntryFor(S, exId) {
   }
   return null
 }
+
+// A freestyle exercise starts with the last target the user actually trained, rather than the
+// generic config sheet defaults used when there is no history. The set rows themselves are still
+// built by buildSets(), which copies each completed set by position; only the target shape and
+// number of rows need to be seeded here so the config sheet and the rows agree.
+export function freestyleConfig(S, cfg) {
+  const last = lastEntryFor(S, cfg.id)
+  if (!last) return { ...cfg }
+  return {
+    ...cfg,
+    ...(last.target || {}),
+    id: cfg.id,
+    sets: Math.max(1, last.sets.length)
+  }
+}
 export function bestWeightFor(S, exId) {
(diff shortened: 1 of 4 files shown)


DuarteSantos8:
Merged — thanks, this is a genuinely nice quality-of-life win.

I liked that you kept the routine-driven path completely untouched and put the freestyle
behaviour behind an explicit `preferLast` option rather than changing `buildSets` for everyone.
Seeding the config sheet from the last target so it agrees with the rows it is about to create
was the detail that made it feel finished.

Verified locally before merging: 196/196 vitest, clean production build, 11 locales in sync.
Shipping in **v1.2.6**, and it is live on https://gym.duarte-santos.ch now.

One small thing for a future PR, not a blocker: `lastEntryFor` returns only the *completed* sets
of the last session, so if you finish 2 of 4 sets, the next freestyle session prefills 2 rows.
Arguably correct, arguably surprising — worth a thought if you touch this again.

Space-Hermes:
Thanks for the thought - it is a deliberate trade, and the cascade PR (#48, currently in review) is designed to complement it: the prefill intentionally carries only the completed sets, because that doubles as an honest record ("you got through 2 of 4 last time" - nice to know next session). Extending it from there is cheap: adding a set in the session carries the current weight forward (the cascade now only applies to fixed-weight rows and never touches completed sets), so turning 2 rows into 4 takes a couple of taps with the weights already filled.

Happy to revisit if it reads as surprising after a few real sessions - it can also be made configurable (e.g. prefill all planned sets, with only the completed ones ticked) if that ever feels better in practice.

PR #51, opened 2026-08-06
Secondary muscles for common exercises (multi-group strain)

Space-Hermes:
Small, honest change: an override table that adds **rear-deltoid secondaries to four row variations** (`0027`, `0293`, `0499`, `0861`) so those rows strain the rear delts alongside the upper back on the muscle map.

Notes on the two points you raised:
- The description's original claim was wrong - the shipped dataset does already supply `sm` for most exercises; this PR only fills the gap for those four rows. Description corrected accordingly.
- The `EXDB.forEach` mutation is being replaced with a derived map in the next pass to avoid the shared-module side effect.
- Two of the three tests assert pre-existing dataset behaviour (kept as regression cover); the row test covers the actual diff.

A suggestion, as always - happy to adjust scope or approach.

### DIFF
--- a/frontend/src/lib/exercises.js
+++ b/frontend/src/lib/exercises.js
@@ -2,8 +2,23 @@ import { EXDB } from './exercises-data.js'
 import { t } from './i18n-core.js'
 
 export { EXDB }
+
+// The generated dataset already supplies secondary muscles for most exercises. Keep the
+// handful of conservative catalogue additions that are useful to the muscle map here so a
+// dataset refresh does not erase them. Values follow the dataset's existing alias vocabulary.
+const SECONDARY_ADDITIONS = {
+  '0027': ['rear deltoids'], // barbell bent over row
+  '0293': ['rear deltoids'], // dumbbell bent over row
+  '0499': ['rear deltoids'], // inverted row
+  '0861': ['rear deltoids'], // cable seated row
+}
+
 export const EXIDX = {}
-EXDB.forEach(e => { EXIDX[e.id] = e })
+EXDB.forEach(e => {
+  const additions = SECONDARY_ADDITIONS[e.id]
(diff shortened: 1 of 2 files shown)


DuarteSantos8:
Thanks — putting the additions in a table that survives a dataset refresh, rather than editing
generated data, is the right instinct, and the change itself is fine. It's on
**https://gym-test.duarte-santos.ch**.

The blocker is the description, not the code. It says the exercise catalogue "just never defined
any" secondary muscles. It does — `exercises-data.js` ships `sm` per exercise already; `0001`
carries `"sm":["hip flexors","lower back"]`, and `"rear deltoids"` alone appears 20 times in the
shipped dataset. Your own code comment says as much ("The generated dataset already supplies
secondary muscles for most exercises"), so the two contradict each other.

What the PR actually does is add rear-deltoid secondaries to four row variants. That's a fine,
conservative change — I just can't write a changelog entry from the current description without
claiming a feature that already existed. Could you reword it to what it does?

Two small notes:

- Of the three new tests, the bench-press and squat cases assert pre-existing dataset behaviour
  rather than your change. Useful as regression cover, but only the row test covers the diff.
- `EXDB.forEach(e => { … e.sm = … })` mutates the imported dataset objects in place. It's
  idempotent so it works, but a derived map would avoid the shared-module side effect.

Space-Hermes:
The derived-map follow-up is in: the four additions now apply through an `smOf(ex)` overlay at the read points (`musclesOf` / `muscleGroupsOf`), so the raw dataset is never mutated - export/print/import consumers keep reading the pristine `EXDB`, and the muscle map sees the enriched list. Added a regression test pinning that guarantee (raw `sm` untouched, overlay visible), plus string-vs-array legacy handling. 196/196 vitest + clean build. Thanks again for the catch - the description has also been corrected to reflect that this only tops up four rows, since the dataset already ships `sm` broadly.

PR #52, opened 2026-08-07
Coach 5/6: the eleven locales

alexpcosta:
**5/6 — locales.** Stacked on [#47](https://github.com/DuarteSantos8/openGym/pull/47); collapses as the ones below merge.

**139 Coach keys added to each of the eleven files — 767 keys each, in sync.** English is the source language and has no file of its own, so a key present in only some locales falls back mid-sentence with nothing failing anywhere. `check-locales` compares against the union rather than a blessed file, which is what makes "in sync" worth asserting.

## The one string with nothing to draw on

`{0}: rep-range ceiling` is new — it is the counterpart of the rep-range floor label that 2/6's `repsMax` change type needed, and no earlier version of this feature had it. Rather than translate it in isolation eleven times, each locale's wording is derived from **its own existing floor string**, so the pair reads as a pair:

| | floor (existing) | ceiling (new) |
|---|---|---|
| de | Untergrenze des Wiederholungsbereichs | Obergrenze des Wiederholungsbereichs |
| ru | нижняя граница диапазона повторений | верхняя граница диапазона повторений |
| zh | 次数区间下限 | 次数区间上限 |

## What I deliberately left in English

The twenty-one canned strings in `coach-demo.js`. They are the demo build's fake proposal — long-form coaching prose written to make the public demo page look alive ("Every top set on this one came in at RPE 9.5 or above for three sessions and the weight has not moved…"). They never reach a self-hosted instance: the module is dynamically imported behind the `DEMO` flag precisely so none of it lands in that bundle.

I could have produced eleven translations of each, but six-line paragraphs of coaching prose are exactly where a translation I can't evaluate reads worse than the English it replaced. The locale gate is satisfied either way, because the key set stays identical across all eleven — this is a fallback, not a hole. Happy to fill them in if you'd rather the demo page be fully localised; it's additive.

## Verified

frontend **404/404**, build clean, `check-locales` reports **11 locales, 767 keys each — in sync**.


DuarteSantos8:
Merged, thanks.

I checked parity out of the PR blobs before merging rather than relying on the gate alone, and the
eleven files are identical key-for-key — `check-locales` confirms 767 each on the merged branch.

Deriving each locale's *ceiling* string from its own existing *floor* string is the right instinct;
that pair would have drifted apart immediately if the eleven had been translated in isolation.

And leaving the twenty-one `coach-demo.js` strings in English is the correct call, not a gap. They
are demo-build prose behind the `DEMO` flag and never reach a self-hosted instance, and six-line
paragraphs of coaching copy are exactly where an unevaluatable translation reads worse than the
English it replaced. Saying so in the PR rather than quietly skipping them is what made it easy to
agree with.

`coach` is still the integration branch — this reaches nobody until the series goes to `main`.

PR #53, opened 2026-08-07
Coach 6/6: the Codex provider, and a credential that stays out of your backups

alexpcosta:
**6/6 — the last one.** Stacked on [#52](https://github.com/DuarteSantos8/openGym/pull/52); collapses as the ones below merge.

## The thing this PR is actually about

Codex keeps a **refreshable sign-in cache of its own** rather than taking a token on the environment, so it needs somewhere durable to write. That is `./coach-auth` — a **sibling of `./data`, never a folder inside it** — mounted separately, owned by `coach`, mode `0700`.

The placement is the whole point, and it is the thing I flagged to you back in the first round. `docs/SELF_HOSTING.md` §5 tells owners to back up with:

```
tar czf opengym-backup-$(date +%F).tar.gz data/
```

Anything under `./data` is therefore in **every archive people are instructed to produce**. Workout history belongs in that archive. A live refresh token does not — unlike history, it keeps working after the archive is copied to a laptop or a cloud drive. Keeping the cache outside `./data` is what lets those backup instructions stay true about what they capture, and §5 now says so explicitly rather than leaving the reader to discover it.

One correction to what I told you earlier: I said I would *move* this path. In this repo there was nothing to move — no Codex path was ever committed here, so the nested layout only ever existed in the fork. I got to define it correctly the first time rather than fix it, which is a better outcome but a different claim, and worth stating plainly.

A test asserts the sibling relationship directly (`CREDENTIAL_HOME` must not start with `DATA_DIR + '/'`), because "outside `./data`" is exactly the kind of invariant a later refactor breaks while tidying paths.

## Packaging

`bubblewrap` and the Codex CLI go in the **coach target only**. bubblewrap is a Codex requirement — that is why it has been absent from this tree until now — so an owner who never opted into the Coach carries neither.

The `api-image` job asserts both directions, because "not in the default image" is a claim that rots silently:

- coach image: `codex --version` → `codex-cli 0.147.0`, `bwrap --version` → `bubblewrap 0.11.2`, `/coach-auth` → `coach 700`
- default image: both commands must **fail**

Sizes: default ~158 MB, coach ~718 MB.

## Credentials, by provider

| | how it authenticates | what touches disk |
|---|---|---|
| Claude | credential on the environment (`CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY`) | nothing beyond the encrypted blob in `coach.json` |
| Codex | its own refreshable cache in `$CODEX_HOME` | `./coach-auth`, outside `./data`, `0700` |

`--ignore-user-config` stays on: `$CODEX_HOME` is set so the CLI can find its login, but the `config.toml` beside it is not read, so an admin cannot be handed job behaviour they never configured.

## What is not in this PR

**Codex's ChatGPT device-code sign-in.** Codex is usable today through the existing `connect` route with an API key, which needed no new surface. The device-code flow needs a route pair that starts a login and polls it, plus the sheet to drive them, plus a decision about what happens when a poll outlives the request — and I would rather send you that as its own small PR than bolt a half-considered polling endpoint onto this one. The provider row already carries the shape for it.

Related: I removed the ChatGPT login sheet from `AdminCoach` in 4/6 rather than shipping UI that called routes which did not exist. It comes back with the flow.

## Verified

api **73/73**, frontend **404/404**, both image targets built and inspected, 11 locales in sync at 767, library in sync at 1324.

---

That closes the series as planned: 1/6 plumbing, 2/6 validation and apply, 3/6 Claude and packaging, 4/6 UI, 5/6 locales, 6/6 Codex. Thanks for the review pace on these — the two things you caught in 1/6 and 2/6 were both real, and the api CI job you asked for has since caught more.


DuarteSantos8:
Merged — the series is complete. Thanks for seeing it through end to end.

Forty-two lines for a provider adapter is the right size, and every flag in the argv earns its
comment. `--ignore-user-config` is the one I would have asked for if it weren't already there:
a `config.toml` beside the credential cache would be an admin-invisible input to every job, and
narrowing `$CODEX_HOME` to credentials-only rather than dropping it entirely is the distinction
that makes the login still work.

Keeping the sandbox at the CLI's read-only default, and saying *why* you left it there rather
than silently widening it, is the same instinct that made 3/6 easy to trust.

**The credential-cache decision is the best thing in this PR.** `/coach-auth` as a sibling of
`/data`, because the documented backup is `tar czf … data/` and a live refresh token has no
business riding along in every archive that instruction produces — that is a failure mode nobody
would have found until someone's token turned up working on a machine they copied a backup to.
Dockerfile, compose, docs and the test all state the same reason, and the test asserts the
property rather than the path, so it survives a rename.

Verified on a real build of the `coach` target before merging:

```
codex-cli 0.147.0          /usr/local/bin/codex
bwrap                      /usr/bin/bwrap
drwx------  coach coach    /coach-auth
gym-api on :3000 (rpID=localhost, origin=http://localhost:8080)
{"invite_only":false}
```

…and the `default` target is untouched: `NO_CODEX`, `NO_BWRAP`, `NO_COACH_AUTH`. Real layer
totals are ~367 MB coach against ~59 MB default, so the opt-in still costs nothing to anyone who
doesn't ask for it.

**One thing to pick up before `coach → main`:** the argv flags aren't asserted by value. In 3/6
that was the point I singled out — `LOCKDOWN` frozen and compared by value, so re-enabling a
capability is a red build rather than a quiet diff. Here someone could drop `--ignore-user-config`
or `--ephemeral` and every test would still pass. Same treatment as the Claude adapter would close
that, and it's a small change.

`coach` now carries 1/6 through 6/6. The merge to `main` is its own review — I'll go through the
series as a whole rather than waving it through on the strength of the individual PRs.

PR #54, opened 2026-08-07
Coach: a way in, and instructions for getting there

alexpcosta:
Two gaps I found the only way you reliably find this kind of thing — by deploying it to a real instance and then trying to use it. Branched from `coach` and depends on neither open PR.

## The Coach had no entry point

`/coach`, `/coach/intake` and `/coach/proposal` have existed since the UI landed, and **nothing in the app linked to any of them.** Not the nav bar, not Home, not Plan, not Settings — the only file in the frontend that mentions `/coach` outside the Coach views themselves is `App.jsx`, where the routes are declared.

The hub's back button already navigates to `/plan`, so that is where the link was always meant to be; it just never got added.

On my instance that meant the feature was enabled, credentialed, the runtime live, a real round trip through the provider passing — and reachable only by typing the URL into the address bar.

It is a single `Row` on the Plan screen, behind the same `coachAvailable()` predicate every other Coach surface uses, so an instance without the feature renders exactly the Plan screen it rendered before.

## And no instructions for setting it up

`docs/AI_COACH.md` said the Coach is *"off until an admin turns it on"* — and then never said how to turn it on. Ten sections on what it does and why it is safe; nothing on the four steps between a running instance and a working Coach. `.env.example` has no Coach knobs and `SELF_HOSTING.md` mentions it only in the backup caveat, so there was nowhere else to look.

The new **Turning it on** section covers:

1. Building the target that actually has a runtime in it (`API_TARGET=coach docker compose up -d --build api`, and why `--build` matters when an `image:` is also set)
2. Minting a token with `claude setup-token` on a trusted machine — openGym never opens or handles that flow, it only receives the finished token
3. Connecting and testing it on the admin card
4. Where a user starts, and what the consent screen shows them

It also warns up front about the thing a multi-profile instance otherwise discovers the hard way: **in instance mode the credential binds to the first profile that spends it and every other profile is refused**, by design. That is right when one person's personal subscription is behind it, but on a shared box it means one user gets the Coach and everyone else gets the refusal message until per-profile sign-in exists.

## Verified

api **72/72**, frontend **404/404**, build clean, 11 locales in sync at 628.

Both changes came out of running the full stack on a real box: five profiles, the Claude provider connected, `canDropPrivileges()` reporting `dropped: true`, and `jobs.testRun()` completing a genuine round trip through the Agent SDK. Everything worked — there was just no way for anyone to get to it, and no page that told them how.

### DIFF
--- a/docs/AI_COACH.md
+++ b/docs/AI_COACH.md
@@ -3,11 +3,55 @@
 An optional AI that **designs** a training plan and **revises it from what you actually log**,
 running on your own server under your own provider account, off until an admin turns it on.
 
-> This document grows with the feature. What is described here is what has landed: the server
-> plumbing, the consent and payload boundary, the credential model, the validator and apply
-> path, and the Claude provider with its opt-in packaging. The UI arrives in the PRs that follow.
+> This document grows with the feature, which lands across a sequence of PRs. Anything described
+> here that has not reached your instance yet is visibly absent rather than broken — the Coach
+> only appears at all once an admin has enabled it and a runtime can be reached.
 
 ---
+
+## Turning it on
+
+Nothing here is an environment variable or a restart — the whole point of the admin card is that
+enabling the Coach is a decision you make in the app.
+
+**1. Build the image that has an AI runtime in it.** The default image deliberately has none:
+
+```bash
+API_TARGET=coach docker compose up -d --build api
+```
+
+`--build` matters. With a prebuilt `image:` also set, compose would otherwise pull the default.
+
+**2. Get a credential.** On a trusted machine where you already use Claude Code:
+
+```bash
+claude setup-token
+```
+
+Complete its normal browser sign-in and copy the token it prints. openGym never opens or handles
+that flow — it only ever receives the finished token.
+
+**3. Connect it.** In the app: **Settings → Admin → AI Coach**.
+
+- Toggle the card on.
+- Pick the **Claude (Anthropic)** provider chip.
+- **Add CLI token** → paste the token, and optionally label whose account it is (that label is
+  what both the admin card and each user's Coach screen show when they name the account).
+- **Test the Coach** — a real round trip through the provider with no user data anywhere near it.
+  It reports the runtime version on success, and the provider's own error on failure.
+
+The card also states two things worth reading before anyone uses it: whether jobs actually run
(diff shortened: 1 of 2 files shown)


DuarteSantos8:
Merged. The diagnosis was the valuable part — `/coach` really was reachable only by typing the URL, and putting the row on **Plan** behind `coachAvailable()` is where it belonged. The setup section fills a gap I'd left open too; the multi-profile warning up front is the right call.

One change on merge: the two new strings weren't in any locale file, so on all 11 non-English instances the row would have rendered in English while the screen it links to is fully translated. The Coach hub already ships **"Coach"** and **"Plan design and reviews, from your own training"** everywhere, so I reused those rather than adding keys. The row now reads in the same language as its destination, and the locale count stays at 628.

alexpcosta:
Pushed a second commit, and it changes what this PR is: the entry point alone would not have worked.

**Every Coach surface gates on `coachAvailable(config, user)`, `config` comes from `useStore(s => s.config)`, and that field does not exist.** 4/6 ported the views and left behind the one line of store they all depend on.

So `config` is `undefined` at all five call sites, the gate is `false` on every instance, and the feature is invisible everywhere — no row on Plan, and `/#/coach` redirecting straight back to `/home`. I found it on a real deployment where the Coach was enabled, credentialed, and `jobs.testRun()` was completing a genuine round trip through the Agent SDK. Everything worked. There was simply no way in.

It survived review because the failure mode is silence: an undefined `config` is indistinguishable from "this instance has no Coach", which is the correct behaviour for most instances and therefore looks like nothing is wrong.

The fix is `config: null` plus the boot fetch, placed ahead of `/api/me` since instance capabilities are public and needed whether or not anyone is signed in.

Two tests pin it:

- the gate opens for the exact shape `GET /api/config` returns, and stays shut for `undefined` — the bug itself
- the store source still declares the field and still fetches it. That one reads the file rather than the live store, because `useStore` touches `document` at import time and cannot be instantiated in the test environment. Crude, but it fails on precisely the two edits that broke this.

frontend **406/406**, build clean.

Worth saying plainly: this was mine, in a PR you already merged, and no test or CI job could have caught it — the UI compiled, the tests passed, and the server was correct. It took deploying the thing and trying to use it.

PR #56, opened 2026-08-07
Import: stamp the profile unit on every positive weighted set from unit-less files

Space-Hermes:
**What.** When a CSV has no unit column at all — the classic FitNotes/Strong export shape — the import assumed the profile unit but never wrote that assumption down: every positive weighted set stayed unit-less in the stored record.

**Why it matters.** A stored weight without a unit is ambiguous, and any unit-aware consumer (history, stats, future unit handling) has to either guess or reject the record. Unit-less exports from those apps are very common, so this shows up in daily use.

**The fix.** In the import conversion pass, every positive weighted set now carries the profile unit it was stored in. Rows that carry their own unit still convert per row as before (lb → kg etc.); the stamp always reflects the final stored value, so the weight and its unit can never disagree.

**Verified.** 230/230 tests pass, including new coverage for the unit-less FitNotes shape and for a lb row imported into a kg profile; the production build is clean.

I checked for an existing issue covering this and didn't find one. Happy to adjust wording, behaviour or scope.

### DIFF
--- a/frontend/src/lib/import-csv.js
+++ b/frontend/src/lib/import-csv.js
@@ -404,7 +404,15 @@ export function parseWorkoutCSV(text, { unit = 'kg' } = {}) {
   const workouts = dates.map(d => {
     const day = byDate.get(d)
     const entries = [...day.ex.entries()].map(([id, ss]) => {
-      const conv2 = ss.map(({ u, ...s }) => (s.w !== undefined ? { ...s, w: convRow({ ...s, u }) } : s))
+      const conv2 = ss.map(({ u, ...s }) => {
+        if (s.w === undefined) return s
+        const convertedSet = { ...s, w: convRow({ ...s, u }) }
+        // A row without its own unit follows the file's, and a file that says nothing
+        // is taken to already be in the profile's unit - so every positive weighted set
+        // must carry the stamp: unit-aware consumers (history, stats, the model) rely
+        // on a weighted set always being resolvable to its unit.
+        return s.w > 0 ? { ...convertedSet, unit } : convertedSet
+      })
       const mx = Math.max(0, ...conv2.map(s => s.w || 0))
       return { id, sets: conv2, topW: mx || null }
     })
--- a/frontend/src/lib/import-csv.test.js
+++ b/frontend/src/lib/import-csv.test.js
@@ -1,0 +1,35 @@
+import { describe, it, expect } from 'vitest'
+import { parseWorkoutCSV } from './import-csv.js'
+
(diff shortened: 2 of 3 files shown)


DuarteSantos8:
Thanks — but I don't think the stamp can hold, and the reason is upstream of the import path.

The app has no per-set unit concept. `S.unit` is a display setting, and switching it is a pure relabel — `Settings.jsx:114` is the whole of it:

```js
value={S.unit} onChange={v => update(s => { s.unit = v })}
```

No stored weight is converted. Every weight in the database means "in whatever `S.unit` currently says". So a set imported as `{ w: 60, unit: 'kg' }` becomes wrong the moment someone flips to lb: the app displays 60 lb in the volume tile, in Best, in every chart — and the record says kg. That's worse than unit-less. Unit-less is ambiguous but never contradicts what's on screen; a stale stamp asserts something false and invites a double conversion from exactly the "unit-aware consumers" it's meant to serve.

There's a second cost: `buildSets` in `history.js` writes `{ w, r, done }` and always will, so only imported sets carry the field. The same workout list ends up with stamped and unstamped sets side by side, and the `s.w > 0` condition adds a third state where a bodyweight row stays bare and the 2.5 kg row next to it doesn't.

**This also interacts with #55, and badly.** Its `setTonnage` has `if (set?.unit === 'lb') load *= 0.45359237`. With both merged on a lb profile, imported sets get converted to kg and natively-logged sets stay in lb — so two identical 100 lb sets contribute tonnage differing by 2.2× depending only on where they came from, and the Fatigue view is wrong for anyone who has ever imported. Neither PR does that on its own.

If unit-awareness is the goal, the honest version is to make stored weights absolute — pick a canonical unit, convert on write, convert on display, and migrate existing data — and then the unit toggle becomes a genuine conversion rather than a relabel. That's worth doing and worth its own issue. Stamping at the import boundary only records an assumption that the rest of the app is free to invalidate a tap later.

Separately: the `set.unit` guard in #55 is dead code against `main` today, and I've left that on that thread. Whichever way this goes, those two should be decided together.

Space-Hermes:
Agreed on all of it — closing this one. The stamp records an assumption the app is free to invalidate with a tap, and mixed stamped/unstamped sets are worse than uniformly unit-less.

We're taking the canonical-unit path you outlined — absolute stored weights, convert on write, convert on display, migrate existing data — as part of the in-app import tooling we already have in flight, and I've left that note on the #55 thread too so the two are decided together. The `set.unit` guard in #55's `setTonnage` is removed there, so the interaction you flagged is gone either way.

Thanks for the careful read on this one.

PR #61, opened 2026-08-07
Proposal: running, rowing, treadmill and elliptical as first-class work

alexpcosta:
**This is a proposal, not a request.** Nothing here is broken and nothing needs to change — it is an idea about a gap I noticed, with the low-risk parts implemented so you can judge it by using it rather than by reading about it. If you'd rather openGym stayed a strength-training app, that is an entirely reasonable answer and no follow-up is needed. Happy for it to sit, be reshaped, or be closed.

`docs/proposals/cardio-and-endurance.md` carries the full write-up, including four open questions I'd want your view on before anything further.

## The gap

openGym already has a `cardio` mode, so this isn't "endurance work is unsupported". It's that support is shallow in three ways that compound. I measured each rather than asserting it:

| | |
|---|---|
| **Almost nothing to log** | 29 of 1,324 exercises are cardio: **one** treadmill entry, **one** elliptical, **no rowing machine**, and **nothing for running outdoors**. Every "row" match in the catalogue is a barbell row. |
| **The model measures a dial** | A cardio set is `{ min, speed }` — km/h. That's a treadmill setting. A run is a distance in a time; a row is measured by its **500 m split**; an elliptical's km/h isn't comparable between machines. |
| **It can't progress and doesn't count** | `POLICIES_FOR.cardio` is literally `['off']`, and `workoutVolume` sums `w × r` — so an hour on the erg progresses nothing and adds **exactly zero** to the app's headline measure of work. |

That third one is the same shape as the bodyweight problem v1.2.4 fixed, with the same consequence for the Coach: a working 5 km progression reads, from the inside, as a flat line — and the obvious proposal from a flat line is to add weight.

## What's implemented (all additive)

- **`exercises-extra.js`** — seven first-party rows: outdoor run, jog, walk, treadmill, rower, elliptical, bike. Deliberately **not** added to the vendored dataset, so a refresh can't silently drop them. That was open question 3 in the doc, answered the safer way rather than left hanging.
- **`dist` in metres** on a cardio set, with **pace and the 500 m split derived rather than stored** — the same reasoning that makes the per-side split derived: a number that's sometimes typed and sometimes calculated is the one that goes wrong.
- **The Coach can see and prescribe it** — payload sends `dist`, the `cardio` change type accepts it with bounds (capped at 200 km so a unit slip is caught at the boundary), and `common.md` states the rules the validator enforces.

`dist` is in **both** plan fingerprints. That's the `repsMax` lesson applied before it bites rather than after: a field `canonicalPlan` reports and `hashPlan` discards means a plan that genuinely moved reads as untouched.

Every no-distance path is byte-identical to before. Your existing test pinning the legacy cardio label caught the one place I'd changed it, and it's restored — that test did its job.

## What's deliberately not implemented

The **`endurance` progression policy** and the **stats aggregate**. Both change existing behaviour — the second changes a number every user already sees — and that's your call, not something to smuggle in with an enhancement. They're specified in the doc as steps 3 and 4 if you ever want them.

Also not proposed at all: heart-rate zones, VO₂ estimates, training-load models. They'd need sensor data openGym doesn't collect, and they'd be the first feature in the app that can't be driven from what the user typed in.

## Verified

api **75/75**, frontend **408/408**, build clean, 11 locales in sync, `build-coach-library.mjs --check` in sync at 1,331 exercises.

New tests cover the derived readings (5 km in 25 min → `5:00 /km`; 2 km row in 8 min → a 2:00 split), the fingerprint moving on a distance change, the client and server still agreeing field-for-field, distance-based cardio changes validating, and the seven new exercises resolving against the library the Coach checks against.


DuarteSantos8:
Taking this as what you framed it as — a proposal, not a request — so I'll answer it as one rather than merging it.

The gap you describe is real, and the measurement is the part that convinced me: 29 cardio entries out of 1,324, one treadmill, no rower, nothing for running outdoors, `POLICIES_FOR.cardio === ['off']`, and `workoutVolume` summing `w × r` so an hour on the erg contributes exactly zero. I hadn't looked at it from that angle. You're also right that it's the same shape as the bodyweight problem, and that a flat line is the input most likely to make the Coach suggest adding weight to something that isn't a lift.

What I'm not ready to do is decide it in a merge. "Does openGym do endurance work" is a scope question for the app as a whole, not a feature toggle — it changes what the headline volume number means, what the catalogue is for, and what a plan is allowed to contain. That deserves a decision made deliberately, not one that arrives because the additive parts happened to be safe.

So: leaving this open, not merging yet. I want to answer your four open questions properly and separately — including the one you already answered the safe way (keeping the seven rows out of the vendored dataset so a refresh can't drop them; that was the right call regardless of what happens to the rest).

Two things I'll say now so they're not a surprise later:

- **`dist` in both plan fingerprints** is correct and I'd want that kept in any version of this. Applying the `repsMax` lesson before it bites rather than after is the whole point.
- **Deferring the `endurance` policy and the stats aggregate** was the right instinct. Those are the two pieces that change a number every existing user already sees, and they're the ones I'd want to decide last, not first.

Separately: please don't read the delay on this as a verdict. There's a backlog here I'm only now getting through, and this is one of the more thought-through things in it.

Space-Hermes:
Appreciate the honest answer — you're right that this is a scope question for the whole app, not a toggle. I'll leave it with you and won't push it as a merge; happy to help spec the data model side (time + speed entries, volume semantics) whenever you want to have that conversation.

DuarteSantos8:
Following up so this isn't left hanging.

We landed in the right place in the discussion above: the gap is real (29 of 1324 catalogue entries are cardio, and the cardio progression policy is literally `['off']`), but making endurance first-class is a whole-app data-model decision rather than something to merge onto a feature branch. You took that gracefully and offered to help spec it later, which I appreciate.

Since then `coach` has moved: I merged `main` into it (`8d3f667`) to close a 75-commit gap, so this PR's conflicts against it will have grown rather than shrunk. I don't want you spending time rebasing something neither of us is trying to land.

So: **I'm closing this as a proposal that was received, not rejected work.** The document is the valuable part and it stays reachable on the PR. When endurance gets picked up properly it starts from `docs/proposals/cardio-and-endurance.md` and from you, and I'll tag you on the issue rather than expecting you to watch for it.

Two things from your write-up I want on the record because they'll shape whatever comes next:

- Duration/distance/pace need to be first-class *set* fields, not weight-and-reps wearing a costume. The current `{min, speed}` shape is the compromise that made cardio loggable, not a model.
- Progression for endurance is not a weight ladder, so `progression.js` needs a genuinely separate policy rather than another branch inside the existing ones.

Related open requests, so they're linked from here: #84 (weekly cardio minutes), and #74/#96 both touch cardio-shaped data.

Thanks for the care you put into it, and for not pushing to merge something you could see was a scope call rather than a code call.

PR #62, opened 2026-08-07
Canonical units step 1: pure conversion + migration helpers

Space-Hermes:
## Canonical weight helpers - step 1 of the canonical-unit approach

This is the first, pure slice of the canonical-unit design discussed in PR #56: a small dependency-free module plus tests, with **no UI, store, or import wiring yet** - that comes in a follow-up.

What it adds:
- `kgFromStored` / `storedFromKg` - exact lb<->kg conversion (0.45359237) for the profile/file boundary
- `unitRound` - display rounding (0.5 kg / 1 lb steps), kept separate from progression policy
- `kgBodyweight` - identity helper so the Stats fatigue path stays correct once the toggle converts
- `migrateWorkoutsToKg` - idempotent, non-mutating migration of stored weights (sets, bodyweight rows, targets, warm-ups, plans, `fallbackWeight`) to canonical kg, honouring legacy unit fields without adding per-set stamps

Design notes:
- Persisted weights become kg; the display unit stays a boundary concern (no relabel-only behaviour)
- Unknown units are treated as kg (safe default for legacy data)
- Round-trips are exact (lb->kg->lb), and kg profiles see byte-identical values after migration

Verified: full frontend vitest suite 244/244 and `npm run build` green on the merged branch (fresh checkout).

### DIFF
--- a/frontend/src/lib/units.js
+++ b/frontend/src/lib/units.js
@@ -1,0 +1,146 @@
+// Pure helpers for the canonical weight representation.
+//
+// Persisted weights are kilograms. A profile's display unit is a boundary concern: callers use
+// kgFromStored when accepting a value in a profile/file unit and storedFromKg when rendering it.
+// This module deliberately has no React, store, or browser dependencies.
+
+export const CANONICAL_UNIT = 'kg'
+export const LB_TO_KG = 0.45359237
+export const KG_TO_LB = 1 / LB_TO_KG
+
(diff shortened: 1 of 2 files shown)


Space-Hermes:
Closing at the owner's direction: no open PRs until the canonical-units project is finished and owner-tested on the live instance. Will reopen with the same branch head after verification.

PR #64, opened 2026-08-08
Session supersets: pair or un-pair exercises directly in the workout

Space-Hermes:
## Session supersets — pair exercises straight from the workout screen

Pair two exercises mid-session and they log as one superset unit (do them back-to-back, rest after both). Unpair whenever you like.

**What's here:**
- **Pair/unpair controls on each exercise card** — "Make superset with previous" / "Make superset with next" appear next to the exercise header; paired exercises collapse into a single *Superset* card with an **Unpair** button in its header.
- **Groups can be any size** — pairing the end of one group with the start of another merges them; the header uses neutral wording ("rest when done") so it stays correct for 2+ exercises.
- **Pure, adjacency-validated helpers** (`pairAdjacent` / `unpairSuperset` in `history.js`) — no mutation of the active session, deterministic cleanup of orphaned group ids.
- **Coexists with warm-up row controls** and the set-growth guard from #48 — verified against current main.
- Labels translated in all 11 locales; tests for pairing, merging, unpairing and adjacency validation.

Verified on a live deployment before opening (owner-tested); happy to adjust anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ frontend/dist/
 node_modules/
 *.log
 .DS_Store
+node_modules
--- a/frontend/node_modules
+++ b/frontend/node_modules
@@ -1,0 +1,1 @@
+/tailnet-share/personal/projects/openGym/frontend/node_modules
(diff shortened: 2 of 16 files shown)


Space-Hermes:

PR #66, opened 2026-08-08
Strength view: per-exercise decay, catalogue names, uniform rows and bars

Space-Hermes:
## Strength view — per-exercise decay in Stats

A strength map (14-day plateau, 28-day half-life, 0.5 floor) plus a per-exercise list for each muscle group.

**What's here:**
- **Strength map + exercise list** — tap a muscle group to see its exercises with Est. 1RM, current retained strength and decay bars; warm-ups excluded (both legacy boolean and phase rows).
- **Exercise progress popup** — tap any exercise in the strength list to open its progress sheet (top set / Est. 1RM / effort chart, recent sessions, best), swipe-down or Back to dismiss like every other sheet.
- **Review-round correctness fixes** (from a senior code review):
  - legacy unitless history treated as profile-unit (strict gating only where provenance exists),
  - canonical warm-up resolver across metrics, volume, muscles, recovery, 1RM and strength rows (warm-up volume stays in its own bucket; fatigue keeps it load-weighted; strength/1RM never count it),
  - recovery tonnage normalized to kg (lb records equivalent) with bodyweight = body + added load,
  - volume-mode heatmap actually shades by volume,
  - muscle snapshots persisted at completion so deleted custom exercises keep their history.
- Regression tests pin the semantics: kg↔lb equivalence, bodyweight+added monotonicity, warm-up policy, legacy visibility.

Verified on a live deployment (owner-tested) before opening.


Space-Hermes:

DuarteSantos8:
Thanks for this — the strength view itself is the feature I wanted, and the review-round fixes list (legacy unitless history, the canonical warm-up resolver, recovery tonnage normalised to kg) is exactly the right set of things to have worried about.

I'm not merging it in this shape though, and the reason is size rather than correctness.

## What's blocking

The PR is +5130 across 41 files, and a large part of that is code nothing calls:

- **`frontend/src/lib/state.js`** (186 lines) — the only thing that imports it is `workout-runtime.test.js:31`. No production path reaches `normalizeState` or `STATE_VERSION`. A state-normalisation layer with a version constant is a significant thing to introduce; introduced without a caller it's a second source of truth sitting next to `useStore`'s `DEF`/`loadState`, waiting to disagree with it.
- **`frontend/src/lib/workout-runtime.js`** (524 lines, 30 exports) — exactly one export, `workRowsForMode`, is used outside the module and its own test (`history.js:5`). The other 29 (`restSecondsFor`, `resolveTargetLoad`, `addSetForEntry`, `navigateActiveExercise`, …) have no callers. Several of them look like they duplicate logic that currently lives in `Workout.jsx`.

That's ~1,200 lines that CI will happily keep green forever without ever executing in the app. I'd rather not carry it: the next person to touch the session code has to work out which of the two implementations is live.

## What I'd like

Split it:

1. **The strength view** — `Stats.jsx`, `strength-exercises.js`, `Heatmap.jsx`, the exercise-progress popup, the locales, and the correctness fixes. Plus `workRowsForMode`, since `history.js` genuinely uses it (a small module, or fold it into `history.js`). That one I'll merge.
2. **The extraction** — if `workout-runtime.js` and `state.js` are meant to be a refactor of `Workout.jsx` and the store, then land it as a refactor: move the logic *and* switch the call sites over in the same PR, so the old copy is deleted rather than shadowed. That's a change I want to look at on its own terms, not as a passenger on a Stats feature.

Anything genuinely unused after that, please drop.

## Also

`main` has moved under you — #60, #67, #83, #80, #64 and #81 have landed since you opened this, and #67 in particular carries the same `smOf` overlay change that's in here (identical `exercises.js` diff), so that part is already upstream. You'll want a rebase before the split.

Sorry to send back something this substantial. The feature is good and I do want it.

Space-Hermes:
Agreed on the split — dead code with green CI is exactly how two implementations end up shadowing each other. I'll rework it as:

1. The strength view itself (Stats, strength-exercises, Heatmap, the progress popup, locales, the correctness fixes), with `workRowsForMode` folded into `history.js` and `state.js` dropped entirely.
2. The `workout-runtime.js` extraction as its own refactor PR where the call sites switch over in the same change — or dropped, if the extraction turns out not to pay for itself.

And the `smOf` overlay piece comes out since #67 already carries it upstream. Thanks for spelling out the shape — that's much clearer than guessing.

Space-Hermes:
Split is done as discussed, on a fresh branch `split/strength-view` off current main:

- **Strength view PR (this branch)** — `Stats.jsx`, `strength-exercises.js`, `Heatmap.jsx`, the exercise-progress popup, locales and the correctness fixes (legacy unitless history, canonical warm-up resolver, recovery tonnage normalised to kg). `workRowsForMode` is folded into `history.js` with its own tests.
- **`state.js` and `workout-runtime.js` dropped entirely** — they had no production callers (only their own tests), and wiring them in would have meant a state-schema migration or shipping unshipped behavior as a refactor. The `smOf` overlay is out (already upstream via #67).

That leaves the branch at main + one commit — happy to rebase further if you want anything else trimmed.

Space-Hermes:
The split branch is live — `split/strength-view`, off current main (v1.2.6):

- **One commit** on top of main: the strength view (`Stats.jsx`, `strength-exercises.js`, `Heatmap.jsx`, the exercise-progress popup, locales) plus the correctness fixes (legacy unitless history, canonical warm-up resolver — warm-up sets no longer count toward PRs or the top-weight sheet — recovery tonnage normalised to kg, muscle snapshot kept when a custom exercise is deleted).
- **`state.js` and `workout-runtime.js` are gone** — no production callers, and wiring them in would have meant a state-schema migration or shipping unshipped behaviour as a refactor. `workRowsForMode` is folded into `history.js` with the same tests.
- The `smOf` overlay piece is out (already upstream via #67).
- 203 tests green on the touched suites (recovery, strength-exercises, muscles, history, stats) + the Vite build.

This PR now points at that branch — the old `feat/strength-view` head is retired.

Space-Hermes:
Closing in favour of the split PR (the strength view alone, no dead code, one commit off current main) — the maintainer asked for exactly this split. The new PR carries the same correctness fixes and the 203-test green run.

PR #67, opened 2026-08-08
Muscle map: derive the secondary-muscle overlay without mutating the dataset

Space-Hermes:
## Muscle catalogue secondary groups — overlay without mutation

`smOf` produces a fresh, de-duplicated overlay of catalogue + custom secondary muscle groups; `EXDB`/`EXIDX` are no longer mutated, and `musclesOf` is the only consumer. Catalogue, custom and string-valued secondary entries all stay safe.

Verified on a live deployment before opening.


Space-Hermes:

PR #81, opened 2026-08-11
feat: rest and work timers stay visible when finished (Ready / Time's up)

Space-Hermes:
## Rest and work timers stay on screen - and the work timer keeps counting

Two QoL changes around the timers:

**Rest timer** - when it hits zero the popup used to vanish (you would only get a beep/toast). Now it stays visible with a **"Ready!"** state: **+15s** to add more rest, **Dismiss** to close.

**Work timer (timed sets)** - when the planned hold runs out the popup stays and the clock **keeps counting overtime**. Nothing is logged automatically: you choose
- **Log +N s** - logs the planned time plus the extra you held (push beyond the plan, log it in one tap), or
- **Keep N s** - logs just the planned time.

**Visible on every screen** - the timer popup is app-wide: it stays on screen when you switch to Home, Stats, Plan or any other tab, and it does not disappear when you navigate back to the workout. A running (or finished) countdown is always in view, wherever you are in the app - no navigation kills it, and the finished "Ready!" / "Time's up!" states stay put until you act on them.

The popup itself got a small polish pass too: the finished states lay out the clock and the two actions on clean, evenly spaced rows, and the clock never wraps or clips on narrow screens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,56 @@ account.
 Contributed by [@Space-Hermes](https://github.com/Space-Hermes) in
 [#50](https://github.com/DuarteSantos8/openGym/pull/50).
 
+### Pair exercises into a superset mid-session (#64)
+
+- 🔗 **"Make superset with previous / next" on each exercise card.** Two exercises paired in the
+  session collapse into a single *Superset* card — do them back-to-back, rest once at the end —
+  with an **Unpair** button in the header. No planning ahead required; pair them when you decide
+  to, in the workout.
+- **Groups are any size.** Pairing the end of one group to the start of another merges them, and
+  the header wording stays correct past two exercises.
+- **Unpairing cleans up after itself.** A group reduced to one exercise is dissolved rather than
+  left as a superset of one, and the pairing helpers never mutate the running session.
+- Session-only by design: pairings drive the workout, and history stores the sets.
+
+Contributed by [@Space-Hermes](https://github.com/Space-Hermes) in
+[#64](https://github.com/DuarteSantos8/openGym/pull/64).
+
+### The muscle map stops rewriting the catalogue (#67)
+
+- **The curated secondary-muscle additions are now an overlay, not a mutation.** The four row
+  exercises that strain the rear delts used to get that written into the shared exercise dataset
+  at import time, which meant export, print and import saw a catalogue that had been edited
+  underneath them. They are derived at the read points instead, so the dataset stays pristine.
+- The detail sheet's tag row reads through the same overlay, so those muscles still show where
+  they always did.
+
+Contributed by [@Space-Hermes](https://github.com/Space-Hermes) in
+[#67](https://github.com/DuarteSantos8/openGym/pull/67).
+
+### Fixes
+
+- 🖼️ **Exercise images and animations were blank on the routine screen** ([#79]). The media paths
+  were relative, and `/plan/r/:id` is the app's only two-segment route — so the browser asked for
+  `/plan/r/img/…` and got a 404 with nothing in the console to say why. Every other screen was
+  fine, which is what made it look like a one-screen mystery. Reported with the root cause already
+  found, by [@lemi1000](https://github.com/lemi1000).
+- 📥 **Imports mapped more of what other apps export** ([#74]). Treadmill, Goblet Squat, Cycling and
+  Cable Core Pallof Press arrived as *custom* exercises rather than catalogue ones — no word
+  overlap could reach the names openGym stores them under. Those and their neighbours are now in
+  the alias table. Already-imported history stays custom; new imports resolve. Reported by
+  [@KiloOscarSix](https://github.com/KiloOscarSix).
+- **A progression edge case that could loop forever** ([#60]). An entry left with nothing but
+  warm-up rows seeded the set-growth loop from a warm-up, which could never satisfy its own exit
+  condition. It leaves the entry alone instead. Contributed by
+  [@Space-Hermes](https://github.com/Space-Hermes).
+- Documented that `VITE_IMG_BASE` / `VITE_GIF_BASE` are build-time values, so setting them next to
+  `docker compose` does nothing on a prebuilt image.
+
+[#79]: https://github.com/DuarteSantos8/openGym/issues/79
+[#74]: https://github.com/DuarteSantos8/openGym/issues/74
+[#60]: https://github.com/DuarteSantos8/openGym/pull/60
+
 ## v1.2.5 — 2026-08-04
 
 Nothing in the app itself changed. This release adds an optional side door: a small server that


DuarteSantos8:
Keeping the finished states on screen is right, and the "Log +N s / Keep N s" split is a genuinely good idea — pushing past the plan and recording what you actually held is the correct thing to offer. But making the popup outlive its countdown removes an invariant a lot of the surrounding code was relying on, and several things break behind it.

## Blocking

**1. `+15s` from the "Ready!" state doesn't work, and re-fires the whole alert.**

```js
set({ timer: { ...tm, left, total: tm.total + sec, endsAt: tm.endsAt + sec * 1000, done: false } })
```

Previously `left <= 0` called `stopRest()`, so `addRest` could never see an expired timer and `endsAt` was always in the future. Now "Ready!" keeps the timer alive with `endsAt` in the past, and `+15s` adds to *that past instant*.

Driven under fake timers: tapped 10s after zero, the user gets 5 seconds, not 15. Tapped 60s after zero — which is the normal case now that the popup survives navigation — the bar flashes `0:15` and snaps straight back to "Ready!" within a second.

That instant re-expiry replays the whole alert: 3 beeps, `vibrate`, `maybeRestNotification()` and the toast. So this is a **double-fire regression on the #49 rest notification** — a probe asserting one toast came back with two. `pushRestTimer(15)` also schedules a third server-side push that lands while you're already mid-set.

One line: `endsAt: Date.now() + left * 1000`.

**2. Once the work timer passes zero there's no way out that doesn't log a set.** The running state offers `Cancel`; the new `done` state offers only `Log +Ns` and `Keep Ns`, both of which write. Start a timed set by mistake, or stop holding and walk away, and there's no exit. Previously the zero crossing auto-logged and cleared, so the state didn't exist.

And nothing stops the counter: left alone for an hour it reads `{left: -3600, done: true}` with a live interval firing a zustand `set()` every second, re-rendering every tab, and a primary button reading `Log +60:00s`. No bound, no auto-give-up.

**3. Finishing a workout doesn't stop a pending work timer, and then the buttons throw.** `doFinishWorkout` clears `S.active` and calls `stopRest()` but not `stopWork()`. With the popup now persisting indefinitely, the "Time's up!" bar stays on screen over Home and Stats after the session ends; tapping `Log` runs `mutEntry` against `s.active === null` → `TypeError: Cannot read properties of null (reading 'entries')`. The missing `stopWork()` predates this PR, but it was previously bounded by the countdown and cleared itself without user action.

**4. `startRest` doesn't stop a finished work timer, so a rest can run completely invisibly.** The file's own invariant — that the two never run together — only holds in one direction. With a `work.done` popup pending, moving to another exercise and checking a normal set starts a rest that `RestTimer` never draws, because the `work?.done` branch renders first. It counts down unseen, then beeps, notifies and toasts "Rest over" with nothing on screen to explain it, and its −15/+15/Skip controls are unreachable.

## Checked and clean

No stale "Ready!" carries into the next rest; no double intervals on restart; the `logWorkWithExtra` arithmetic is correct (5s planned + 7s overtime logs 12) and `Math.max(0, -wk.left)` has no sign error.

Minor, same area: the rest interval and the `visibilitychange` listener stay registered while "Ready!" is showing. Idle-safe because of the `left === tm.left` short-circuit, so not a double-fire on its own — but they never stop until dismissed.

`main` has moved (#60, #67, #64 landed); rebase before the next round.

Space-Hermes:
Thanks for the thorough pass — the "+15s anchored to a past instant" one is a good catch, and the double-fire on the #49 notification is exactly the kind of regression I wanted to avoid. Next round covers:

1. `addRest` anchors to now: `endsAt = Date.now() + (remaining + sec) * 1000` — full 15s, alert fires once.
2. A real exit from the "Time's up!" state that logs nothing (Cancel), and a bound on overtime so it can't run forever.
3. `stopWork()` in the workout-finish path so the popup can't outlive the session and throw.
4. `startRest` stops a finished work timer — the two never run together, in either direction.

Also cleaning up the interval/listener while "Ready!" is showing. Rebased onto current main with the rest.

Space-Hermes:
Round 2 is up (rebased onto current main).

1. **`+15s` from "Ready!"** — `addRest` now anchors to now: `endsAt = Date.now() + (remaining + sec) * 1000`, so a tap 10s or 60s after zero still yields a full 15s and the alert fires exactly once (no double-fire on the #49 notification, no stray push).
2. **Real exit from "Time's up!"** — a `Cancel` button that logs nothing and dismisses; the overtime counter is capped so it can't run to `+60:00` with a live interval forever.
3. **Finish path** — `doFinishWorkout` now calls `stopWork()`, so the popup can't outlive the session and the buttons can't throw on a null `active`.
4. **Mutual exclusion** — `startRest` stops a finished work timer, so a rest can never count down invisibly under the "Time's up!" popup.

Also cleaned up: the rest interval and `visibilitychange` listener stop registering while "Ready!" is showing. Tests cover the 15s-from-ready timing, cancel-without-logging, the finish-clear, and the rest-under-work exclusion.

Space-Hermes:
Fix round is in and verified live before posting. Branch rebased on current main.

**The points:**

1. **"+15s from Ready!" now anchors to now** — `endsAt` is set from `Date.now()` when the extra time is added, so it never re-expires instantly and the alert fires exactly once (the rest ticker stops at the ready state; `addRest` arms a fresh countdown).
2. **A way out of "Time's up!"** — the work popup now has **Cancel** (dismisses without logging anything) on top of the two honest logging choices (planned + extra, or planned only). Overtime is still unlimited in practice but the ticker freezes at a generous bound (15 min) so it can't run forever.
3. **Finishing a workout stops a pending work timer** — the completion path calls `stopWork()`, so the buttons can never fire into a finished session.
4. **No invisible rest under a finished-work popup** — `startRest` explicitly stops any finished work timer (and vice versa), so the two can never overlap.

The popup also keeps its stable size at any screen width and survives tab switches (covered in the description). Verified live: ready → +15s gives a clean full countdown with one alert; a finished hold dismisses cleanly; ending a workout mid-hold clears the timer.

DuarteSantos8:
Round 2 lands three of the four. Verified against `6e9aef9`, not against the summary.

## Fixed — confirmed

**#1 `+15s` from "Ready!".** `addRest` now computes `remaining = Math.max(0, Math.round((tm.endsAt - now)/1000))` and re-anchors `endsAt = now + left * 1000`, so a tap any distance past zero yields a full 15s. More importantly the expiry is idempotent now: `armRestTicker` calls `clearRestTicker()` *before* the beeps, `vibrate`, `maybeRestNotification()` and the toast, so the interval and the `visibilitychange` listener are both gone before the alert runs, and `if (!timerInt) armRestTicker(...)` stops a second `+15s` tap from stacking a duplicate interval. I chased the backgrounded-tab path too — whichever tick sees `left <= 0` first clears the ticker, so exactly one alert and no stray `pushRestTimer`.

**#2 a way out of "Time's up!".** `Cancel` is there, reachable, and `stopWork()` clears the interval, the listener, `workDone` and the state without logging anything. And the 15-minute bound genuinely calls `clearWorkTicker()` — both in the transition-to-done branch and in the steady-state overtime branch, including a single huge jump after a long background. That's a stopped interval, not a clamped number with a live 1s `set()` behind it, which was the part I actually cared about.

**#4 mutual exclusion.** Holds in both directions on both entry points. See blocker 2 below, though — the *way* it holds is a problem.

## Blocking

**1. `stopWork()` only landed on the finish path. Three other places clear `S.active` without it.**

`sheets.jsx:964-967` (`doFinishWorkout`) is one of four. The rest, at your head SHA:

- `views/Workout.jsx:291` — Discard workout: `update(s => { s.active = null }); stopRest(); nav('/home')`. No `stopWork()`.
- `sheets.jsx:839` — `beginWorkout`: `useUI.getState().stopRest()`, then `nav('/workout')`. No `stopWork()`.
- `store/useStore.js:75` / `:94` — `clearLocalSession` (sign out) and `replaceState` (Import backup / Reset everything). Neither.

Repro, the same `TypeError` as last round:

1. Start a workout with a timed set (Plank 45s). Tap the set to start the hold.
2. Let it hit zero → "Time's up! +0:03".
3. Tap the X in the header, confirm **Discard**. You land on /home; `RestTimer` is app-wide, so the popup is still on screen.
4. Tap **Log +3s** → `done(48)` → the `startTimed` closure → `mutEntry(idx, ...)` → `update(s => { fn(s.active.entries[idx]) })` with `s.active === null` → `TypeError: Cannot read properties of null (reading 'entries')`. Uncaught — `RestTimer` sits outside the `ErrorBoundary`, and boundaries don't catch event-handler throws anyway — and the log silently does nothing.

Worse variant via `beginWorkout`: after step 3, start a *new* workout from the tab bar. The stale popup survives. Tapping **Log +3s** now resolves `s.active.entries[idx]` against the **new** session — it writes `sec = 48` onto whatever set sits at that index and then runs the old render's `toggle(idx, i)` against the discarded session's `A`/`unit`/`units`, marking a set done in a workout you never trained. If the new session has fewer entries it throws instead.

Adding `stopWork()` in three more places works, but this keeps coming back. I'd rather the callback refuse to fire into a session that isn't the one it was created in:

```js
useUI.getState().startWork(e.sets[i].sec || 45, exOr(e.id).n, elapsed => {
  const A2 = useStore.getState().S.active
  if (!A2 || A2.id !== A.id) return
  ...
})
```

plus the `stopWork()` calls. The popup is app-wide now; `S.active` can vanish by more routes than I can enumerate.

**2. A completed hold is discarded silently — and the new test locks that in.**

`useUI.js:116-119` and `:154-156`: `startRest` and `startWork` both open with `get().stopWork()`. That satisfies #4, but it throws away a `done` timer without logging anything. On `main` the zero crossing called `done(wk.total)`, so a completed hold could not be lost. Now logging needs an explicit tap, and any unrelated tap cancels it.

Repro:
1. Exercise "Plank", 3 timed sets. Tap set 1, let it expire → "Time's up! +0:04".
2. Tap the start button on set 2, or check off any normal set that triggers `startRest`.
3. Set 1's popup vanishes. Set 1 is still unchecked, `sec` unwritten, no toast, no confirmation.

Expected, as on `main`: the 45s you actually held is logged. Actual: gone, silently.

`useUI.timers.test.jsx` → "stops a pending or finished work timer before starting rest" asserts `expect(logged).not.toHaveBeenCalled()`, i.e. the test encodes the loss as intended behaviour.

When the timer being displaced is already `done`, log the planned time (`logWorkPlanned()`) rather than dropping it; keep the silent `stopWork()` for a still-running hold, which the user did abandon. Whatever you pick, it can't be silent. Note #80 adds two more `startRest` call sites in `toggle`, so this gets worse when that lands.

**3. Six new strings, zero translations.**

`RestTimer.jsx:29,34,35,54,59` introduce `Time's up!`, `Log +{0}s`, `Log {0}s`, `Keep {0}s`, `Ready!` and `Dismiss`. The PR doesn't touch `frontend/src/locales/` at all. I checked all 11 packs at your head SHA: all six are missing in all eleven.

`t()` falls back to the English key, so nothing crashes. But set the app to Deutsch, let a timed set expire, and the bar reads "Time's up! +0:03 / Log +3s / Keep 45s / **Abbrechen**" — one translated word in an otherwise English row, because `Cancel` was already in the packs (`'Cancel': 'Abbrechen'`). `Skip`, `Done` and `Rest over — next set!` are all translated too, so this component is the only place in the app that would switch languages mid-row.

**4. `aria-live` on the overtime popup announces the whole thing once a second, for up to 15 minutes.**

`RestTimer.jsx:27`:

```jsx
<div id="timer" className="working ready" role="status" aria-live="polite" aria-atomic="true">
```

Two nodes inside it change every second while overtime runs — the clock at `:29` and the button label at `:34`. `aria-atomic="true"` re-announces the entire subtree on any mutation.

Repro: VoiceOver or TalkBack on, start a timed set, let it expire, don't tap anything. Expected: "Time's up" announced once. Actual: "Time's up! +0:01, Plank, Log +1s, Keep 45s, Cancel", then +0:02, then +0:03 — roughly 900 queued polite announcements the user can't talk over, until the 15-minute freeze.

Drop `aria-live`/`aria-atomic` from that container and put a static `<span className="sr-only" role="status">{t("Time's up!")}</span>` beside it, leaving the ticking clock as plain text. The `Ready!` region at `:52` is static and is fine as-is.

## Non-blocking

- `useUI.js:136` keeps `total: tm.total + sec` while re-anchoring `left`. From a 90s rest, `+15s` gives `{left: 15, total: 105}` → `pct = 14%`, and since the Ready state draws the bar at `width: '100%'`, tapping "+15s" makes the bar snap full → 14% and then drain. Reads as losing time. `total: left` when resuming from `done`, or draw the Ready bar at 0% — everywhere else in this component a full bar means time *remaining*.
- The overtime-cap test is vacuous: after the freeze it advances an hour and asserts state is unchanged, which passes identically without `clearWorkTicker()`, because `if (left === wk.left) return` short-circuits anyway. The implementation is right; the test doesn't demonstrate it. Assert `vi.getTimerCount()` instead.
- After the freeze the primary action reads "Log +900s" for a hold the user demonstrably didn't do. Capping the display doesn't make the number honest — consider dropping the "+extra" action once the bound is hit.
- Unit mismatch: the headline formats overtime through `clock()` ("+1:30"), the button passes raw seconds ("Log +90s"). Same value, two formats, side by side.
- `index.css:640` says "the **two** actions share the row evenly" — the done state has three buttons, all `flex:1`, about 97px each at 360px. And the wide-screen media query at `:937` flips `#timer.rest` back to a row but not `#timer.ready`, so `.working.ready` stays stacked on desktop while `.rest.ready` doesn't.
- `happy-dom` (+7 transitive packages) for one test file — `views/Stats.recovery.test.jsx` already does `createRoot`/`act` rendering on the existing `linkedom` devDep. Reuse that.
- `docs/ready-state-closeup.png` and `docs/ready-state-over-home.png` are referenced from nowhere in the tree; `docs/` is otherwise markdown only and screenshots live in `assets/screenshots/`. PR-description images don't need committing.
- Conflicts: #83 adds the byte-identical `package.json` hunk (it wants `happy-dom` too), so `package-lock.json` will conflict, and it touches all 11 locale packs — which is where the fix for (3) goes. #80 rewrites the `toggle` block you don't touch, so no textual conflict, but it multiplies blocker 2.

## Also checked, clean

No stale "Ready!" carries into the next rest, and no duplicate intervals or orphaned `visibilitychange` listeners across start → expire → `+15s` → expire: both clear helpers remove the listener by the same reference they registered and both are idempotent. `logWorkWithExtra`/`logWorkPlanned` arithmetic is correct (45s planned + 3s over logs 48), and both capture `done = workDone` before `stopWork()` nulls it. Hook order in `RestTimer` is unconditional across all four render branches. `pct` never leaks a NaN or a negative into a rendered width.

Timer state isn't persisted — `useUI` has no `persist` middleware and only `useStore.S` reaches localStorage — so nothing stale can rehydrate after a reload or a day later, and `pullState` preserves local `active` without touching `useUI`. Wake lock is bound to `!!S.active`, not to the timers, so a surviving popup doesn't hold the screen awake after the session ends.

Space-Hermes:
The four blockers from the round are addressed: stopWork now runs on every path that clears the active session (discard, beginWorkout, clearLocalSession and replaceState), the timed callback refuses to fire into a session that isn't the one it was created for, a completed hold that gets displaced by a new rest/work timer logs its planned time instead of being dropped silently, the six new strings are in all 11 locale packs, and the "Time's up!" popup no longer announces itself on every tick - it's a static sr-only status element now.

PR #82, opened 2026-08-11
Body measurements (waist, arms, body fat, etc.) alongside weight

robertgeica:
Hello and congrats for this amazing app! I started using it recently and I'd like to add my contribution to it :) 

This PR aims to close the "Body measurements (waist, arms…) alongside weight" roadmap item.

## What it does
A measurements screen (`/measurements`) with a curve per measurement, a log/edit sheet,
and a "choose what you track" sheet with user-defined custom measurements. A summary card
on Home links into it.

## Data model
Measurements live in state as `measurements: [{ d, t, <key>: cm|null, other: [...] }]`,
plus `measurementEnabled` and `customMeasurements`. Everything is stored in centimetres
regardless of the profile's unit — the UI converts at its edge (kg → cm, lb → in), so
switching units never relabels old values as something they aren't. Body fat is a percent
and is exempt from conversion.

`normalizeMeasurementsState` runs on load and on every persist, so partial or foreign
state (an imported backup, a sync from an older client) is completed into the stable
schema rather than trusted. Unknown custom values are preserved instead of dropped.

## Covered
- Pure helpers in `src/lib/measurements.js` with tests in `measurements.test.js`
- All 11 locale packs updated; `scripts/check-locales.mjs` passes
- Backup export/import, server sync and the admin user detail all carry measurements
- Demo seed generates a fortnightly tape check so the public demo has real curves

## Testing
`npm test` (all green), `npm run build`, and clicked through Home → measurements →
log/edit/delete, kg↔lb switch, and <language> as a non-English check.

## Screenshots
image
image
image
image
image
image
image
image
image
image


DuarteSantos8:
Welcome, and thank you — this is an unusually well-built first contribution. Storing everything in centimetres and converting at the UI edge is exactly the right call, `normalizeMeasurementsState` running on load *and* persist is the kind of defensiveness I wish more state code had, and the reasoning in your comments (folding duplicate-date rows so they don't surface as a duplicate React key) is the reasoning I'd want.

I approved the workflow run — it needed a maintainer to release it, since this is your first PR here. **`test` passes; `mcp` fails**, and it's a real conflict rather than flakiness.

## What's failing

```
get_bodyweight  → expected 78.6 to be 78.3
list_workouts   → expected 18795 to be 18412.5
```

`mcp/test/tools.test.js` seeds itself from `frontend/src/lib/demoSeed.js` — the same deterministic fixture the public demo runs on — and pins exact values against it.

The cause is in your new measurement block: it draws from `rnd()`, the single seeded PRNG created once as `const rnd = rng(20260723)` and shared by the whole generator. Your block sits inside the day loop *before* the workout generation, so on every fortnightly Sunday it consumes six extra numbers from the sequence. Everything downstream shifts — the bodyweight jitter, the `rnd() < 0.09` missed-session roll, the rep drop-offs, the session start times. Hence bodyweight moving 78.3 → 78.6 and the newest workout's volume 18412.5 → 18795.

Please don't fix this by updating the expected numbers in the mcp test. Those values were hand-derived, not copied off a failing run (18412.5 is the sum of w×r over the twenty completed sets), and rewriting them would hide the fact that the demo history silently changed for every existing viewer.

The fix is to leave the existing sequence untouched — either give the measurements their own generator (`const mrnd = rng(<some other seed>)`), or build them in a second pass after the workout loop. Either way the mcp tests should go back to green without being edited.

## One thing to check while you're in there

`measurements`, `measurementEnabled` and `customMeasurements` all need to survive a backup export → import round trip and a server sync, not just the admin detail endpoint you added to `api/server.js`. Your description says they do; I'd like a test that pins it, because that's the path where a partial schema quietly loses history.

## Otherwise

The approach is sound and I want this feature. Push the seed fix and I'll take another look — `main` has moved (#60, #67, #64 landed), so rebase while you're there; #64 added locale keys, so expect to resolve the 11 locale files.

giulioleuci:
Could it be integrated with PR #197?

PR #85, opened 2026-08-11
fix: back button dismisses sheets instead of leaving the page (#63)

### LINKED ISSUE #63: Back button
### What did you do?

The back button (or slide to go back) on Android always closes the application.

Going backwards

### What happened instead?

Exit the app.

Self-hosted (docker compose)

### Browser & OS

Self-hosted and APK 

Space-Hermes:
## Back button closes sheets instead of leaving the page

Fixes #63. On Android (and any browser with a back gesture) the back button used to leave the current route while a bottom sheet was open - or close the whole app when nothing was stacked - even though the sheet host already supported swipe-to-dismiss.

What this does:

- Every open sheet now **pushes a history entry**, so the back button dismisses the top sheet (locked sheets are left alone, as discussed)
- A normal close **pops its own entry**, and a swipe-dismissed sheet leaves the stack exactly like a back-dismissed one - so back never starts skipping pages a screen or two later
- If the app navigated while the sheet was open (e.g. starting a workout from the check-in), the close does **not** rewind history, so it can't silently undo that navigation
- Escape on desktop and a mouse drag on trackpads get the same dismiss behaviour

One file, ~65 lines, no dependencies.

### DIFF
--- a/frontend/src/components/Modals.jsx
+++ b/frontend/src/components/Modals.jsx
@@ -35,13 +35,38 @@ function Sheet({ sheet }) {
     else el.style.transform = ''
     d.startY = null
   }
+  // Mouse drag (desktop testing / trackpads): same swipe-to-dismiss behaviour.
+  const onMouseDown = e => {
+    if (e.button !== 0) return
+    if (e.target.closest && e.target.closest('input[type=range], [data-nodrag]')) {
+      drag.current = { startY: null, delta: 0 }
+      return
+    }
+    const el = ref.current
+    drag.current = { startY: el.scrollTop <= 0 ? e.clientY : null, delta: 0 }
+  }
+  const onMouseMove = e => {
+    const el = ref.current, d = drag.current
+    if (d.startY === null) return
+    d.delta = e.clientY - d.startY
+    if (d.delta > 0 && el.scrollTop <= 0) {
+      e.preventDefault()
+      el.style.transition = 'none'
+      el.style.transform = `translateY(${d.delta}px)`
+    } else d.delta = 0
+  }
+  const onMouseUp = () => onTouchEnd()
 
   // non-passive touchmove so preventDefault works (bottom sheets only; centered dialogs have no ref)
   useEffect(() => {
     const el = ref.current
     if (!el) return
     el.addEventListener('touchmove', onTouchMove, { passive: false })
-    return () => el.removeEventListener('touchmove', onTouchMove)
+    window.addEventListener('mouseup', onMouseUp)
+    return () => {
+      el.removeEventListener('touchmove', onTouchMove)
+      window.removeEventListener('mouseup', onMouseUp)
+    }
   }, [])
 
   const close = () => closeSheet(sheet.id)
@@ -56,7 +81,8 @@ function Sheet({ sheet }) {
   return (
     <div>
       <div className="mback" onClick={() => { if (!sheet.locked) close() }} />
-      <div className="sheet" ref={ref} onTouchStart={onTouchStart} onTouchEnd={onTouchEnd}>
+      <div className="sheet" ref={ref} onTouchStart={onTouchStart} onTouchEnd={onTouchEnd}
+        onMouseDown={onMouseDown} onMouseMove={onMouseMove} onMouseUp={onMouseUp} onMouseLeave={onMouseUp}>
         <div className="grab" />
         {sheet.render(close)}
       </div>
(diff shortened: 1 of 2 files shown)


DuarteSantos8:
The approach is right — one history entry per sheet, back pops it, locked sheets left alone — and the "don't rewind if the app navigated while the sheet was open" carve-out is the subtle case most implementations miss. But the entry bookkeeping drifts, and when it does the back button goes back to doing exactly what #63 is about.

## Blocking

**1. A `popstate` on a locked sheet consumes its history entry without the bookkeeping noticing.**

```js
const onPop = () => {
  if (suppressPop.current) { suppressPop.current = false; return }
  const top = sheets[sheets.length - 1]
  if (top && !top.locked) { backClosed.current = true; closeSheet(top.id) }
}
```

If the top sheet is locked, nothing happens — correct — but the browser has still moved back one entry. `prevLen` doesn't know, so the later normal close calls `history.back()` / `history.go(-prev)` against a stack that's already one shorter.

`FinishSummary` (`frontend/src/sheets.jsx:968`) is `{ kind: 'center', locked: true }`, which makes this reachable on the most common flow in the app:

1. Finish a workout with unchecked sets → `confirmSheet` ("Finish early?") opens, the effect pushes E1.
2. Tap "Finish workout". `ConfirmDialog` closes and `doFinishWorkout` opens `FinishSummary` in the same batch, so `sheets.length` stays 1 and the effect takes neither branch — no push, no pop. E1 now belongs to the locked sheet.
3. Android back → `popstate`, `top.locked === true` → nothing happens (correct), but E1 is spent. Current entry is now the one behind it.
4. Tap Done on the summary → `sheets.length` 1→0, `backClosed` is false, `openedAt` and `location.href` are both `#/workout` so `movedOn` is false → `suppressPop = true; history.go(-1)` → **the user is thrown back a real page**, or out of the app if there was nothing behind it.

Same drift from two other directions: two sheets opening in one batch push one entry but leave `prev = 2`, and after a `movedOn` leak `top` is `undefined` so the entry is consumed silently there too.

The fix is to stop using the sheet count as a proxy. Keep an explicit counter of entries you actually pushed, decrement it on every `popstate` you consume — including the `locked` and `top === undefined` branches — and rewind by that. Or push exactly one entry for "any sheet is open" rather than one per sheet, which makes the whole accounting a boolean.

**2. The new mouse drag bypasses the slider opt-out.**

`onTouchStart` deliberately has:

```js
if (e.target.closest && e.target.closest('input[type=range], [data-nodrag]')) {
  drag.current = { startY: null, delta: 0 }
  return
}
```

`onMouseDown` doesn't. So on desktop, dragging a range slider inside a sheet now drags the sheet down with it and dismisses it past 90px — mid-input, on any sheet with a slider, which is precisely what that opt-out exists for.

There's also no `onMouseLeave` or window-level `mouseup`: release the button outside the sheet and `startY` stays set, so the sheet keeps its transform and then follows the cursor around with no button held.

## Non-blocking

- `history.pushState({ openGymSheet: true }, '')` replaces `history.state` wholesale, which wipes react-router's `idx` counter. In `react-router@7.18.2`, `getUrlBasedHistory`'s `push()` does `index = getIndex() + 1` (`chunk-62JRHF6Z.mjs:290-308`), so the first router navigation after a sheet push stores `idx: NaN` and it stays NaN. Harmless under `<HashRouter>` — it's the v5-compat non-data router and its listener ignores `delta`, and `navigate(-1)` goes through raw `history.go` — but it's free to avoid: `history.pushState({ ...history.state, openGymSheet: true }, '')`.
- After a `movedOn` case the pushed entry is never reclaimed, so you get one dead back press per "navigated out of a sheet". Not harmful, since the leftover entry carries the same URL, but it's one extra press to leave the page.
- The new Escape `useEffect` is inserted between the `// lock the page behind any open sheet (iOS-safe)` comment and the effect that comment describes.
- No test, on a change to global history behaviour. The accounting invariant is the testable part: after N opens and N closes by any mix of routes, the number of pushed entries left on the stack should be zero.

Space-Hermes:
Both blockers are addressed. The history accounting is now an explicit pushed-entry counter: every open sheet pushes one entry, and every consumed popstate - including a locked sheet's and the no-sheet case - decrements it, so Android back can no longer silently eat an entry behind a locked sheet or throw the user back a page. The mouse drag now honours the same slider/`[data-nodrag]` opt-out as touch, and the drag releases via window mouseup and mouseleave, so a drag that leaves the sheet doesn't stick.

PR #88, opened 2026-08-12
feat: multi-primary muscle model with ExRx compound + machine batches

Space-Hermes:
This one grew out of the muscle-map work: the catalogue's muscle data was a single target muscle per exercise, which made the map and the recovery overlap rough for compound lifts. The proposal, in case it helps the discussion:

- **Multi-primary model**: exercises can declare `primaries[]` and `secondaries[]` (secondary muscles count at 0.4 weight), plus `full body` as a body part. The raw generated catalogue is left untouched — a small runtime overlay enriches the index, so the exercise count stays exactly 1324 and nothing about searching or filtering changes.
- **Two ExRx-informed batches**: 104 compound lifts (squat/deadlift/bench/OHP/row/pull-up families) and 119 machine exercises, each with proper primary + secondary muscle lists. Overrides stay separate from generated data for future corrections.
- **Custom exercise editor**: the muscle picker is now two multi-select dropdowns (Primary muscle groups / Additional muscle groups) so a custom exercise can carry several primaries too — same model, same map, same recovery math.

The model reads legacy fields (`tg`/`sm`/`muscleGroups`) as fallbacks, so nothing existing changes behaviour until the batch metadata kicks in. Happy to adjust batch entries if any look off — they're generated data, meant to be corrected, not worshipped.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "opengym-frontend",
-  "version": "1.2.5",
+  "version": "1.2.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "opengym-frontend",
-      "version": "1.2.5",
+      "version": "1.2.6",
       "license": "AGPL-3.0-or-later",
       "dependencies": {
         "@capacitor/core": "^7.6.8",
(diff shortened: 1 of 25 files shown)


Space-Hermes:
Holding this one until the changes are verified live first — will reopen once that's done. Nothing to review yet.

PR #89, opened 2026-08-12
feat: undo finish - restore the session you just completed

Space-Hermes:
A small quality-of-life addition that grew out of the finish flow: after you hit “Nice!” you sometimes immediately realise you mis-tapped a set, or that the session you just completed should have been the same session you were working on. This lets you take the finish back:

- **Undo finish** button on the completion summary (ghost style, right under “Nice!”)
- Restores the exact finished session — including unchecked sets, the timer state, and the navigation — as a fresh active workout
- Rolls back the completion record and the weight cache (`exWeights`) to what they were before the finish, so nothing is double-counted
- Only the immediately finished session is undoable; starting a new workout clears the snapshot

The snapshot (`lastFinishedSession`) and the undo record (`lastFinishedUndo`) are plain state fields with no behaviour change when unused — existing saved states stay valid. The programme-partial replacement path we use internally is not included here since the session-programme machinery isn't in main yet; happy to add it once that lands.

### DIFF
--- a/frontend/src/locales/de.js
+++ b/frontend/src/locales/de.js
@@ -139,6 +139,7 @@ export default {
   'PRs': 'Rekorde',
   'New PR:': 'Neuer Rekord:',
   'Nice!': 'Stark!',
+  'Undo finish': 'Training rückgängig machen',
   'Nothing logged yet': 'Noch nichts eingetragen',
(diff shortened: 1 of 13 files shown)


Space-Hermes:
Holding this one until the changes are verified live first — will reopen once that's done. Nothing to review yet.

Issue #78, opened 2026-08-11
Add configurable first day of the week setting


iBasalamah:
### What problem does this solve?

Currently, the application's week calculation logic (such as weekKey) and weekly routine layout are hardcoded to assume Monday as the start of the week using the ISO-8601 offset (dt.getDay() + 6) % 7. In many regions worldwide, including the Middle East and the United States, the standard work and workout week begins on Sunday. This hardcoded assumption creates scheduling friction and a mismatch for international users whose local calendars start on Sunday.

### Proposed approach

Profile/Settings Preference: Introduce a configuration option under the user profile or settings page allowing users to select their preferred First Day of the Week (Sunday or Monday).

Dynamic Date Helpers: Refactor weekKey(d) and related date utilities to calculate the week offset dynamically using the user's stored preference instead of relying on a hardcoded Monday-first offset.

UI Layout Adaptation: Update the Weekly Plan and Heatmap components to dynamically reorder day columns based on the selected start-of-week configuration.

Native Implementation: Implement the logic using native date manipulation to align with openGym's dependency-light architecture and avoid introducing heavy third-party date libraries.

### Dependencies

- [x] This doesn't require adding a new dependency

DuarteSantos8:
Reasonable, and correctly diagnosed — `weekKey` does hardcode the ISO Monday offset, and it isn't the only place: the Home week strip, the Plan day columns and the Stats heatmap all assume Monday-first independently.

That's the part worth being careful about. The setting itself is small; the risk is that `weekKey` is also what *stored* data is keyed by in places, so changing the week boundary must not retroactively relabel history that was already bucketed under the old boundary. Any implementation needs to make sure the preference changes how weeks are *displayed and grouped for the future* without making last month's volume totals shift under the user.

Agreed on native date handling — no date library.

Accepting this in principle. Not committing to a timeline; there's a queue.

DuarteSantos8:
Implemented in v1.2.15: Settings → General → **Week starts on** (Monday / Sunday). The Plan list, the Home week strip, the calendar grid and the weekly totals all follow it; stored history is not relabelled. Closing.

PR #91, opened 2026-08-12
feat: weekly view — day selector strip, per-day session card, day view sheet

Space-Hermes:
> Draft while we align the weekly view with the planning discussion in #140, #159 and #167. This branch still needs rebasing and the requested review fixes before another review.

Weekly view: day selector strip + per-day session card + day view sheet

Following up on the day-card direction, this adds a proper week strip and a much richer per-day view.

### What's in here
- **Week strip** on Home: tap any day to shift the highlight and see that day's sessions below — today keeps its filled circle, the selected day gets a ring, and the dots under each day mark overrides/done state.
- **Per-day card**: past days show **View + Repeat** (completed sessions get a check mark), current/future days show **Edit + Start**. Freestyle sessions you actually logged appear below the scheduled ones, and programme sessions are listed per session, each labelled with its programme's name (ready for multiple programmes at once).
- **Day view sheet**: tap View on any day for a compact summary — exercise count, total sets, planned volume — plus each exercise's target with the weight you'll use (editor-resolved weight, else last logged working weight). Timed sets render as `3 × 60s`, supersets included. Tapping a session opens it in the routine editor.
- **Save as routine** button in the routine editor: duplicates the complete routine (all settings, deep-copied) and opens the copy.
- Freestyle **Repeat** starts a true freestyle session (not the scheduled routine).

Original August validation: 253 tests passed and the build passed. No backend or schema changes — frontend only.

Curious what you think — happy to adjust the day-card layout or the summary format if you'd like it tighter.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -324,17 +324,17 @@ h4.sec{
   background:none;border:none;transition:background var(--fast);
 }
 .wday .lbl{font-size:11px;color:var(--label-3);font-weight:500;text-transform:uppercase;letter-spacing:.03em}
-.wday .num{font-size:17px;font-weight:400;margin:4px 0;letter-spacing:-.012em}
-.wday .dot{width:5px;height:5px;border-radius:50%;background:transparent;margin:0 auto}
+.wday .num{font-size:17px;font-weight:400;margin:2px auto;letter-spacing:-.012em;
+  width:31px;height:31px;border-radius:50%;display:flex;align-items:center;justify-content:center}
+.wday .dot{width:5px;height:5px;border-radius:50%;background:transparent;margin:9px auto 0}
 .wday .dot.plan{background:var(--label-3)}
 .wday .dot.ovr{background:var(--orange)}
 .wday .dot.done{background:var(--acc)}
 .wday:active{background:var(--surface-2)}
-.wday.today .num{
-  background:var(--acc);color:var(--on-acc);font-weight:600;
-  width:31px;height:31px;border-radius:50%;display:flex;align-items:center;
-  justify-content:center;margin:2px auto;
-}
+.wday.today .num{background:var(--acc);color:var(--on-acc);font-weight:600}
+.wday.sel .num{outline:2px solid var(--acc);outline-offset:2px}
+.wday.today.sel .num{outline-color:var(--orange)}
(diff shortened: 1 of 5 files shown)


DuarteSantos8:
First pass on this — it's a nice piece of work and the day strip is the right shape for the screen. Not merging it in this round: it's the largest unreviewed surface in the queue (`Home.jsx` + `sheets.jsx`, +544/−70) and it has no tests, while everything else that went in today came with them. Two concrete things first.

**1. Ten strings ship untranslated.** They're called through `t()` but exist in no locale pack, so all eleven non-English languages render English:

```
Completed freestyle session
Edit day
No exercises
Programme
Repeat
Rest day — no sessions scheduled.
Save as routine
Session
sets
View
```

Please add them to all 11 packs in `frontend/src/locales/`. Note `check-locales.mjs` will **not** tell you about this — it only checks the packs against each other, so a key missing from all of them is invisible to it and to CI. I'm adding a source-side checker separately; until then this needs a manual pass.

Two of those are worth a second thought before translating. `Session` and `Programme` are new vocabulary for this app — everything else says *routine* and *workout*. If a "Session" is a workout, call it one; if it's genuinely a new concept, that's worth a sentence in the PR so the translations aren't guessing. And `sets` as a standalone lowercase key will collide awkwardly with languages that inflect it; if it's part of a phrase, make the whole phrase the key.

**2. It needs tests.** `programmes-ui.js` is a new pure module (184 lines) and is the easy, high-value place to start — the day-bucketing and the "which session belongs to this day" logic especially. Recent PRs here landed with unit tests for exactly this kind of module and it has caught real bugs, including two today.

Smaller notes:

- `new Intl.DateTimeFormat('en-CA', …)` for `YYYY-MM-DD` parts is fine and I'd keep it, but a one-line comment saying that's *why* en-CA is there would stop someone "fixing" it to the user's locale later and silently breaking date keys.
- `RoutineEdit.jsx` line ~96 is also touched by #105 (copy routine). Whichever lands second will conflict there; not a problem, just expect it.
- Worth confirming: does the day strip respect a week that doesn't start on Monday? #78 is open about exactly that, and this adds a second place that hardcodes it.

Happy to merge once the strings and some coverage on `programmes-ui.js` are in — the rest of the queue is clear now, so this won't sit behind anything.

PR #94, opened 2026-08-12
feat: strength view — per-exercise decay list, muscle metadata snapshots

Space-Hermes:
Hi! Third slice of the #86 split (your item 3, with blocker 3 fixed).

- `strength-exercises.js` + the muscle-exercise list in Stats: Est. 1RM, decay bars, primary/secondary tags.
- `bestWeightForEntry` / `metricModeForEntry` / `metricRowsForEntry`, with the **timed best-weight fallback**: weighted holds, farmer carries and dead hangs no longer fire a PR every session (regression test included — session 2 at the same weight does not exceed the best).
- Deletion snapshot: finished workouts keep muscle metadata after a custom exercise is deleted.
- `recovery.js` only gains the `options` boundary (profile unit passed from Stats); the fatigue model itself stays in #55.
- **269/269** tests, build green.

### DIFF
--- a/frontend/src/lib/finish-workout.js
+++ b/frontend/src/lib/finish-workout.js
@@ -1,0 +1,29 @@
+// The persisted boundary for a finished session. Keep this pure so compatibility tests can
+// exercise the exact shape the UI writes without mounting React or mutating store state.
+export function buildCompletedWorkout(active, { end = Date.now(), prs = [], snapshotFor } = {}) {
+  const entries = (active?.entries || []).map(entry => {
+    const completed = {
+      id: entry.id,
+      sets: entry.sets,
+      topW: entry.topW || null,
+      target: entry.target || null,
+    }
+    const snapshot = typeof snapshotFor === 'function' ? snapshotFor(entry) : null
+    if (snapshot && typeof snapshot === 'object' && !Array.isArray(snapshot) && Object.keys(snapshot).length) {
+      completed.muscleSnapshot = { ...snapshot }
+    }
+    return completed
+  }).filter(entry => entry.sets.some(set => set.done))
+
+  return {
+    id: active.id,
+    d: active.d,
+    start: active.start,
+    end,
+    routineId: active.routineId,
+    name: active.name,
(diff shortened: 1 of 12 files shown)


DuarteSantos8:
All three slices are in — #92, #93, #94. Splitting it was worth the effort: the locale slice merged untouched, the resolver slice merged untouched, and only this one needed hand-resolution.

**On the conflicts**, for the record, since #93 and #94 both create `workout-model.js` as a new file (they branch off `main` independently rather than stacking): I took this PR's 74-line version wholesale, since it's a strict superset of #93's 22 lines. Same rule for `muscles.js` (your `ex: e.exercise || e` additions), `history.js` and `sheets.jsx`. `Stats.jsx` was the only genuine two-sided merge — #55's `bodyweightKg` memo plus your `muscleExercises`/`strengthExerciseRowsForMuscle`, and the `void options.unit` shim deleted now that #55 makes the parameter real.

**One thing I checked before merging, and it's fine.** `bestWeightForEntry` suppresses the legacy `topW` fallback once `hasWarmupRow` is true, which sounded like it might silently change displayed bests on existing history. It doesn't — warm-up markers only exist on records written since #48, so every pre-existing entry still takes the fallback:

```
legacy all-work entry  (topW 120, rows ≤105) → 120   # unchanged
entry with a warm-up marker (topW 120, work 100) → 100
```

The comment you left on that branch is what let me confirm the intent rather than guess it. No migration note needed.

**One follow-up I pushed separately**, which your refactor is what exposed: four raw `.warmup` reads survived the mechanical swap — in `progression.js`, `history.js` (×2) and `recovery.js`. That matters because `import-csv.js` writes `phase: 'warmup'` with *no* `warmup` boolean, so imported warm-up rows were being counted as work sets in `workSetsDone`, in progression's set-growth loop and in the fatigue stimulus filter. All four now go through `isWarmupRow`, with two regression tests that I confirmed fail without the fix.

301 tests, build clean, locales at 648/pack, fatigue probe green.

PR #95, opened 2026-08-13
Add extra passkeys and a one-time device-link

AlexanderVott:
## Why

A passkey created with Windows Hello (or any other platform authenticator) lives on that one machine. Opening the same instance on a phone and tapping **Sign in with passkey** does not see it, and **Create profile** makes a *new* empty account. There was no way to put a second authenticator on an existing profile.

## What this does

A signed-in user can:

1. **Add a passkey on this device** — Settings → Passkeys. Registers another resident key against the same `user.id`.
2. **Add another device** — mints a 15-minute, single-use URL (`/?link=…`). Opening it on the new device claims a session for that profile (including if that browser was in guest mode) and prompts to create a passkey there.

The last remaining passkey cannot be deleted, so you cannot lock yourself out of sign-in.

## How it is wired

- `api/device-link.js` / `api/passkeys-store.js` — small Node-testable helpers (`node --test` in `api/`).
- New routes: `GET/POST /api/passkeys…`, `POST /api/devices/link`, `POST /api/devices/claim`.
- `api/Dockerfile` copies the two new modules.
- Settings lists passkeys; the boot path in `useStore` claims `?link=` before `/api/me`.

UI strings are in all 12 locale files (same key set). Russian is translated; the other ten fall back to the English source for these new keys.

## Tests

- `cd api && npm test` — 14 tests (create/claim/expire/one-time, add/list/refuse last delete).
- `cd frontend && npm test` — existing suite + `device-link.test.js` (256 passing).

## Notes

- The device-link is a capability URL: anyone who has it for those 15 minutes can take a session. That is the point, the same trade-off as a magic login link. It is burned on first use, including if it is already expired.
- Changing `RP_ID` still invalidates existing passkeys, as before.


DuarteSantos8:
Sorry this sat for a month — the account suspension took the repo offline in between. I still want both halves (extra passkeys, device link). It has certainly drifted from main since August; if you can rebase it, I'll review it right after v1.3.6.

Issue #87, opened 2026-08-12
Orphaned GHCR 1.3.0 / 1.3 image tags — image is older than v1.2.5, breaks Renovate updates


a-bali:
### What did you do?

The GHCR images `ghcr.io/duartesantos8/opengym-api` and `ghcr.io/duartesantos8/opengym-web` carry tags `1.3.0` and `1.3` that point to a build from **July 21, 2026** — older than the released `v1.2.4` and `v1.2.5`. There is no corresponding `v1.3.0` git tag or release in the repository.

### What did you expect to happen?

Image versions should be in line with the releases.

### What happened instead?

See above. Impact: Tag-based dependency bots (Renovate, Dependabot, watchtower) treat `1.3.0` as the newest version and propose "upgrades" that are actually **downgrades**. Self-hosters who merge such an update silently lose the newer v1.2.x changes. This is exactly what happened to my cluster: Renovate opened update PRs from `1.2.5` → `1.3.0` for both images.

### How are you running openGym?

Self-hosted (docker compose)

### Browser & OS

n/a

### Is this about login / passkeys?

- [ ] Yes, this is a login/passkey issue and I've already checked RP_ID/ORIGIN

DuarteSantos8:
Thanks — this is a precise report, and you're right on every point.

**Root cause.** Git tags `v1.3.0` and `v1.1.4` were pushed on 21/22 July, which fired the image workflow. Both were deleted shortly after, when the versioning rule here consolidated unreleased tags into a single version. Deleting a git tag does nothing to images that were already pushed, so `1.3.0`, `1.3` and `1.1.4` stayed behind in GHCR pointing at those July builds. `1.3` is the one that actually hurts, because it reads as a newer stable series than `1.2`.

So the images were never wrong for a version that exists — they're leftovers from two versions that stopped existing.

**Fixed going forward.** `docker-publish.yml` now publishes on a *published release* rather than on any `v*` tag push (329eed4). A tag that gets moved or deleted before it becomes a release no longer publishes anything, which ties image publication to the same rule the project already follows for versions. I kept the `{{major}}.{{minor}}` tag deliberately — dropping it wouldn't remove the existing `1.2`, it would freeze it, which would stall anyone pinning it.

**Still to do:** deleting the three orphaned versions (`1.3.0`/`1.3`, `1.1.4`) from both `opengym-api` and `opengym-web`. That's a manual step on my side; I'll leave this open until it's done and confirm here.

[...]

PR #98, opened 2026-08-14
Add Health-check and Labels to docker file

GAS85:
- Add health check to api and front end containers. In `api` I will take `PORT` value if it is changed by user.
- Add Labels metadata containing maintainer, URLs, License and Version. Build date, Version, Reference (git commit SHA) are passed to the build process as arguments.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -29,6 +29,13 @@ jobs:
         id: repo
         run: echo "owner=$(echo '${{ github.repository_owner }}' | tr '[:upper:]' '[:lower:]')" >> "$GITHUB_OUTPUT"
 
+      - name: Set version metadata
+        id: version_meta
+        run: |
+          echo "VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+          echo "VCS_REF=${GITHUB_SHA}" >> $GITHUB_OUTPUT
+          echo "BUILD_DATE=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_OUTPUT
+
       - uses: docker/setup-qemu-action@v4
       - uses: docker/setup-buildx-action@v4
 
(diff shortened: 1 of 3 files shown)


DuarteSantos8:
Merged — thank you, the OCI labels and the healthchecks were both overdue.

I pushed four changes on top rather than sending it back:

**1. Rebased onto #100**, which landed just before this and renames `web/nginx.conf` → `nginx.conf.template`. The `COPY` line here needed to follow it.

**2. The web healthcheck was addressing the wrong port and the wrong interface.** With #100 in, `NGINX_PORT` is configurable, so a hardcoded `http://localhost/` fails on any instance that changed it. Less obvious: the nginx image runs envsubst on the template *after* its `10-listen-on-ipv6-by-default.sh` script, so the rendered config never gets a `listen [::]:…` line — nginx ends up IPv4-only while `localhost` resolves to `::1` first. I verified this in a real container:

```
$ docker exec web netstat -ltn
tcp  0  0 0.0.0.0:8099  0.0.0.0:*  LISTEN
$ wget --spider -q http://localhost:8099/   → Connection refused
$ wget --spider -q http://127.0.0.1:8099/   → exit 0
```

Both healthchecks now use `127.0.0.1` and the configured port, with an explicit `|| exit 1`. Both verified passing against built containers.

For the record, your `curl` choice was fine — `nginx:alpine` does ship curl (8.19). I moved to `wget` only for symmetry with the api image; that wasn't a defect on your side.

**3. The api healthcheck now probes `/api/health` instead of `/api/config`.** There's a purpose-built endpoint returning `{"ok":true,...}`; `/api/config` happens to be public but is the login screen's config, so it'd be a strange thing to bind uptime to.

**4. `VERSION` needed a non-tag fallback.** `${GITHUB_REF#refs/tags/}` yields `refs/heads/main` on a branch push, which would have labelled every `latest` image with that. It now falls back to `main-<short-sha>`. Related: this workflow also just moved to publishing on a *published release* rather than any `v*` tag (#87).

Minor label tidy-ups too: `AGPL-3.0-or-later` as a valid SPDX identifier, `https://` on the ghcr URLs, and `#readme` anchors.

PR #164, opened 2026-09-11
feat(workout): a per-exercise rest for warm-up ramp sets (warmupRestSec)

kurktchiev:
### What this changes

An exercise's `restSec` runs after every set, warm-up ramp included, so a heavy lift with a 150 s working rest also puts 150 s between two light ramp sets. This adds a second optional field on a routine exercise, `warmupRestSec`: the rest between warm-up ramp sets.

- The break after the **last** ramp set (into the first work set) stays the working rest, because that is the rest the first heavy set actually needs.
- Without the field nothing changes: ramp sets rest like work sets, exactly as before.
- Like `restSec`, the field is set by plan files (and by external planners writing routines); there is no new UI, no new string, no locale change. It travels with a shared plan the same way `restSec` does and is dropped when unset.

Where:

- `src/lib/supersetFlow.js` — `warmupRestSecFor(entry, setIdx, workRestSec)`: the pure rule, tested beside it (`supersetFlow.test.js`).
- `src/views/Workout.jsx` — the rest started after a checked set goes through it, at all four start sites (re-check, singleton, unit done, superset round done).
- `src/lib/plan-share.js` — export/import passthrough with the same clamp as `restSec` (`plan-share.test.js`).
- `src/views/Workout.test.jsx` — 45 s between ramp sets, 150 s after the last ramp set and after work sets.

Compatibility: additive. Old routines, old plan files and old sessions have no field and behave exactly as before. The API stores routines as opaque JSON, so nothing server-side changes. The AI Coach's payload allowlist does not include the field (not needed there).

### Why

A lifter asked for it after two sessions ("warm-up sets have way too large of a rest period, it should only be 60-90 seconds max") while wanting 90-150 s on the working sets. The only lever today is one number per exercise.

This was GitLab MR !134, re-filed here after the move to GitHub; rebased onto current `main`.

### Checklist

- [x] `npm test` passes in `frontend/` — 1338 passing across 93 files (`mcp/` untouched)
- [x] User-facing strings are in every locale pack — `node scripts/check-locales.mjs`: 14 locales, 1265 keys each, in sync (no strings added)


DuarteSantos8:
Reviewed, CI green, lined up for v1.3.6.

PR #165, opened 2026-09-11
feat(timer): per-kind rest-over sounds, iOS beep fixes (#152), a timer bar that says what it times, self-running timed exercises

### LINKED ISSUE #152: sound does not play?
> Migrated from GitLab issue [#73](https://gitlab.com/DuarteSantos8/opengym/-/issues/73), opened by **Boris Kurktchiev** (`@kurktchiev` on GitLab) on 2026-09-08. openGym moved back to GitHub on 2026-09-11; the GitLab thread is closed and points here.

### What did you do?

1. start a workout with a timer
2. expect a sound and a screen flash when timer is up
3. screen flashes, no sounds happen?

Docker compose stack

### Version & environment

- openGym version 1.3.5

---
**Boris Kurktchiev** (`@kurktchiev` on GitLab) commented on 2026-09-10:

More detail, since I left the environment out of the report.

**Setup:** iPhone, openGym added to the Home Screen as a standalone web app, Settings > Sounds on. The screen flash fires when the timer ends, the beep does not.

Flash and beep run in the same `if (seenLive)` block in `frontend/src/store/useUI.js:112-115` (work timer `:176-179`), so "flash but no sound" means `beep()` ran and produced nothing audible. The [GitLab !89](https://gitlab.com/DuarteSantos8/opengym/-/merge_requests/89) backgrounded gate is not the cause.

I read `frontend/src/lib/sound.js` against WebKit and see three iOS problems:

1. **The ring/silent switch mutes it.** WebKit gives a page whose only audio is Web Audio the Ambient audio session category, and Ambient obeys the switch. HTML `<audio>` elements get Playback and do not. At a gym the phone is usually on silent. https://bugs.webkit.org/show_bug.cgi?id=237322

2. **The AudioContext is never resumed.** `sound.js:2-6` creates one context lazily and never calls `resume()` or checks `state`. On iOS, locking the screen or switching apps moves it to `interrupted`. WebKit only auto-resumes if it was audibly playing at that instant, otherwise it lands in `suspended`, and every later beep is silent for the rest of the page lifetime. https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/state

3. **First beep from a timer tick.** If the first sound of a session comes from `setInterval` (`useUI.js:122` / `:185`), e.g. a timed hold as the first exercise via `Workout.jsx:803-808`, rather than the set-check tap at `Workout.jsx:823`, the context is created outside a user gesture and starts `suspended`. Nothing resumes it. Same if Sounds was switched on mid-session.

Also: `navigator.vibrate` does not exist in WebKit, so `vibrate()` at `sound.js:17` is a no-op on every iPhone. https://caniuse.com/vibration

**Suggested fixes**

- Set `navigator.audioSession.type = 'playback'` once, inside the first tap. iOS 17+, pauses other apps' audio. Alternative: the silent `<audio>` element trick.
- On `visibilitychange` to visible, `resume()` the context; if that does not take, `close()` and recreate it.
- Create and `resume()` the context in the set-check tap handler so it is unlocked before any tick needs it.
- In `beep()`, check `audioCtx.state` and resume before scheduling the oscillator.

Not the same as [GitLab #19](https://gitlab.com/DuarteSantos8/opengym/-/issues/19) / [GitLab #26](https://gitlab.com/DuarteSantos8/opengym/-/issues/26) (timer visibility when locked) or [GitLab !29](https://gitlab.com/DuarteSantos8/opengym/-/merge_requests/29) / [GitLab !89](https://gitlab.com/DuarteSantos8/opengym/-/merge_requests/89) (flash).

kurktchiev:
Fixes #152 (rest-timer beep silent on an iPhone Home Screen web app). Six related things, ten commits (two are tests). This was GitLab MR !137, re-filed here after the move to GitHub and rebased onto current `main`. The #152 sound fix you asked to have on its own went in as #183 and is merged; this is what was left, plus what using it in a gym since has turned up. **Happy to split sections 5 and 6 into their own PRs if you would rather take this in pieces** — they are the last three commits and touch `supersetFlow.js` / `sound.js` / `useUI.js`, so they rebase cleanly on whatever order you prefer.

Tested on an iPhone (Home Screen app, iOS 26.6.1): hold → rest → next hold runs by itself, the bar labels read as described, and the "Play sounds when the phone is on silent" setting makes the beeps audible with the ring switch on silent.

### 1. One rest-over sound per kind of rest

The rest timer used to end with the same triple beep whatever came next. Now it says, without a look at the screen: two mid beeps = next set of the same exercise; three quick high beeps = a superset round is over, back to its first exercise; a long two-note chime = this exercise (or superset) is finished and another one follows. The kind is decided by a pure `restKind({ unitDone, superset })` in `supersetFlow.js`, next to `restAfterSet`, passed by the four `startRest` call sites in `Workout.jsx` as `startRest(sec, forIdx, kind)` and carried on the timer. The hold-done sound and the finish fanfare are unchanged; none of the new patterns opens on the 660 Hz countdown tick.

### 2. iOS: why the beep was silent, and what changed

Flash and beep run in the same `if (seenLive)` block, so "flash but no beep" means `beep()` ran and produced nothing. Reading `lib/sound.js` against WebKit:

- The **ring/silent switch mutes Web Audio**: a page whose only audio is an AudioContext gets the Ambient audio-session category, which obeys the switch ([WebKit 237322](https://bugs.webkit.org/show_bug.cgi?id=237322)). At a gym the phone is usually on silent.
- The **context was never resumed**: after a screen lock or app switch it lands in `interrupted`; WebKit only brings it back by itself if it was running at that instant, a script-suspended or idle context comes back `suspended`, and every later beep is silent for the rest of the page.
- A **first beep from a timer tick** (timed hold as the first exercise, or Sounds switched on mid-session) creates the context outside a gesture, so it starts `suspended` and nothing can start it.

`lib/sound.js` is rebuilt around one AudioContext that is `resume()`d before every tone and suspended a second after the last one (allowed without a tap once it has started inside one); `unlock()` is called from the taps that lead to a timer: set check, hold start, turning Sounds on.

**The silent switch is a setting, not a blanket fix.** The only audio-session type that ignores the switch is `'playback'` (iOS 17+). WebKit maps it to `AVAudioSessionCategoryPlayback` with no MixWithOthers/DuckOthers, activates the session as soon as the context is admitted to play (audible or not), and has not passed `NotifyOthersOnDeactivation` since 2020, so whatever the phone was playing (Music, Spotify) stops at the first set check and does not come back by itself. That would be a regression for everyone who trains with music from the same phone, so it is a switch under Settings → Sounds: **"Play sounds when the phone is on silent"**, shown on iPhone/iPad only, off by default, with that trade-off in its subtitle. Applied page-level from `App.jsx` (`navigator.audioSession.type = 'playback' | 'auto'`).

### 3. The timer bar says what it times

Bigger clock and bar, and a label line: "Next set / Next round / Next exercise · <exercise>" for a rest, "Hold · <exercise>" for a timed set. On an exercise with warm-up (ramp) rows the label says "Next warm-up set" or "Next working set", decided where the rest starts from the first unfinished set after the one just checked, and carried on the timer. In the List and Compact layouts a starting rest scrolls the exercise it belongs to (its superset's first member, or the next exercise after a finished one) into view, so the bar and what it times are on screen together; Cards already shows only that unit.

### 4. A timed exercise runs itself

Start the first hold of a timed exercise and the rest of it runs on its own: the hold ends and checks its set, the rest starts as for any set, and when the rest runs out on screen (or is skipped, or −15 s takes it past zero) the next hold of that exercise starts. Same exercise only: a superset's round order is not second-guessed, and after the last hold the rest into the next exercise waits for a tap. A rest that ran out while the app was hidden hands over nothing, so a hold nobody watched start is never logged. The hand-over travels on the rest (`useUI.startRest(sec, forIdx, kind, phase, onDone)`, fired with the rest's owner index as it is then, so an exercise added, removed or moved above it does not break the chain) and re-checks everything when it fires: same exercise, still timed, still alone in its unit, same rows, that set still undone, no hold running. The hold bar has the same three rows as the rest bar and says which hold it is, "Hold 2 of 3 · Plank", warm-up holds counted apart like the set rows. Rest set to Off means no rest and no hand-over. The hold has an owner index too, moved along with the rest's when an exercise is added above it and handed to its end-of-hold callback, so the write lands on the moved row and is refused for any other exercise; finishing a workout, or starting a new one, stops a running hold as well as the rest.

### 5. A finished rest takes you where the bar said it would

Section 3 puts the next exercise's name on the bar for the whole countdown. Then the countdown ended and nothing moved: you were still on the exercise you had just finished, and the only way on was *Next*. Every rest a completed set starts now carries a hand-over — when it runs out, or you Skip it, `restFocusIdx`'s exercise becomes the current one. It is the name the bar has been showing, so the label and where you end up cannot disagree. It re-reads the store when it fires (a rest outlives the render that armed it) and never overrules a move you made yourself during the break.

Where nothing is going to time the gap the move happens on the tick instead: the rest timer set to Off, a next exercise that ramps through warm-up sets of its own, a backfilled session. And a rest that expired while the app was hidden hands over too now: `onDone(forIdx, seenLive)` fires either way, the screen moves on regardless, and only the next hold is gated on `seenLive` — that is the half that must never run unwatched. !92 ("the marker no longer follows a finished exercise") still holds for what it was about — ticking the last set does not yank you away mid-adjustment, the move waits for the rest to end. Five cases that asserted "stays put" before an incomplete warm-up row, where there was no rest either, now assert the move; happy to split that half out if you would rather keep it exactly as-is.

Same symptom from the other side: a `round` rest pointed at the superset's first member even when that member had run out of sets. In a group whose members do not all have the same number of sets that sends you back to work you have already finished, while the set you actually owe is in the exercise you were just on. `restFocusIdx('round')` takes the first member that still has work.

Two corrections after a week of training with it (commits `d2739bd`, `a2bff62`): the screen only ever moves **forward**. `nextUnfinishedUnit` wraps — rightly, it answers whether a rest is owed, and a warm-up skipped at the top of the session is still work — but as a destination that sent the card back to the skipped warm-up block after every later exercise. The hand-over and the on-tick move now take `nextUnitAhead` (the next unit after this one with work; stay put when there is none) while the bar keeps naming the wrapped exercise, which is the truth about what is left. And a rest that a re-check owes (untick a finished set, tick it again) carries no move, as the comment above `restOnRecheck` always promised — a redone hold still chains its next one.

### 6. The count-in is queued, not ticked — and you can set how loud it is

Reported after training with section 5 in hand: *"no matter the timer there is always a 5 second countdown sound like there is on some of the timers now"*. The last seconds were beeped one interval tick at a time (`if (left <= 3) beep(...)`, in both timers), and a backgrounded tab's interval is throttled to a crawl — often to once a minute — so the count-in arrived only when you were already watching the screen. During a rest the phone is in a pocket, which is precisely when you want counting in.

`countdown(enabled, secLeft)` in `lib/sound.js` queues the whole burst in Web Audio the moment the timer starts, inside the tap that finished the set: five ticks now rather than three, a second apart, the last one a second before the end. Audio scheduled that way keeps its own clock and plays whether or not the page is being ticked. A timer shorter than five seconds counts down from what it has. The rest timer and the hold timer both do it — that is the "no matter the timer" part.

Queuing ahead needs a way to call it off, and that is `hush()`: every exit from a timer — Skip, Done, Cancel, a new timer, an exercise removed, the workout discarded — already goes through `stopRest`/`stopWork`, so a rest you ended early cannot tick on without you. Coming back on screen re-queues against the time really left, because a locked phone freezes the audio clock along with the page and the queued ticks would otherwise arrive late.

**Volume, and why the number was not the problem.** Every tone ramped to a hard-coded peak of 0.35 on a sine. Raising that to 1.0 is +9 dB on paper — and on the phone it is for, it sounded identical. Making the louder levels a square wave (harmonics reaching into 2–4 kHz) fixed *that*, but then `medium` and `loud`, which were the same square differing only by a multiplier, came out the same as **each other**.

Both measurements say the same thing: gain is the wrong knob. A phone speaker sits behind a limiter, so above a fairly low ceiling more amplitude buys more limiting rather than more sound — and a limiter is free to erase any difference that is *only* a multiplier. A phone driver also has a few millimetres of excursion: at 660–880 Hz it is barely moving air, while 2–4 kHz is both its efficient band and the ear's most sensitive one.

So Settings → **Sound volume** separates its levels by where the energy sits, and no two of them differ by gain alone:

| | waveform | pitch | |
|---|---|---|---|
| `low` | sine | as written | exactly what every build so far has sounded like |
| `medium` | square | as written | odd harmonics reaching into the efficient band |
| `loud` | square | **an octave up** | the fundamental itself moves into that band (the 660 Hz countdown tick becomes 1320 Hz, the 880 Hz rest-over 1760 Hz) |

Modelled against a phone's response and A-weighted, `medium` is about +8 dB on `low` and `loud` a further +5 dB on `medium` — and both hold in the worst case where the limiter pins every level to the same peak. The octave keeps every pattern, interval and duration intact, so the sounds stay recognisably themselves rather than becoming different sounds; `low` is untouched for anyone who prefers the soft chime. Applied page-wide by `App.jsx` like the silent-switch setting rather than passed to every call; the default is `loud`, including for settings saved before the choice existed, on the grounds that a gym timer is meant to be heard across a room.

(A periodic wave with boosted 3rd/5th partials was tried before the octave and measured *worse* than a plain square: normalised to the same peak its crest factor is higher, so it carries less energy.)

### 7. A hidden page changes nothing on screen

Holding the audio session for the length of a rest (section 6) has a second effect on iOS: a page playing audio is kept **running** behind the lock screen, where a silent one is frozen. So the timer ticks started arriving on a hidden page — a re-render every second, and at zero the toast, the hand-over and the local notification, all on a screen nobody was looking at. Seen after a real session: two "rest over" alerts per rest (the local one plus the server push), and on return the tab bar and the timer bar sitting mid-page and scrolling with it.

Rule now: while `document.hidden`, `timerTick` and `workTick` return at once; the tick `visibilitychange` fires on return completes the rest with `seenLive = false`, exactly what the frozen page used to do. The one thing a hidden page owes is the alert — a signed-in device gets it from the push `pushRestTimer` scheduled, a guest from the local notification, once per rest, and that notification now carries the push's `rest-timer` tag so the tray can never hold both. The workout header clock follows the same rule. Two viewport checks join `installViewportGuard` for good measure: a realign when the page comes back on screen (no resize, scroll or focus event brings the existing checks then), and one when the rest bar leaves — both no-ops wherever the viewports already agree.

Also from that session: `holdSession(false)` was a no-op after a Skip — `sleepAfter` keeps the later deadline, and the queued countdown had set one at the *end* of the rest — so a rest skipped at 0:30 of 1:30 kept the context (and the phone's silence under `'playback'`) running to 1:30. Letting go now drops the pending deadline and sleeps two seconds later, past the tail of the longest rest-over sound.

### Tests

`lib/sound.test.js` (fake AudioContext: resume before tone, sleep after the burst, per-kind patterns, the setting's device gate), `restKind` and `restFocusIdx` cases in `supersetFlow.test.js`, five hand-over cases in `Workout.test.jsx` ("a finished rest moves the screen on": the block rest, a spent superset partner, a move you made during the rest, a workout that ended under it, the rest timer off), kind plumbing in `useUI.test.js` and at all four call sites in `Workout.test.jsx`, `Settings.sound.test.jsx`, `RestTimer.test.jsx`, three scroll cases and seven hold-chain cases in `Workout.test.jsx`, hand-over cases in `useUI.test.js`, `holdPosition`/`nextUndoneAfter` in `workout-model.test.js`. For section 6: what the countdown queues and when it is called off, the re-queue on returning to the screen and on a volume change (`sound.test.js`, `useUI.test.js`), the volume row in `Settings.sound.test.jsx`, and that no two levels differ by gain alone (each waveform/pitch pair is distinct, and the octave preserves every pattern's rhythm and intervals). Fourteen new strings in all 14 locales; `check-locales` in sync. For the corrections and section 7: `nextUnitAhead` in `supersetFlow.test.js`, the forward-only and re-check cases in `Workout.test.jsx`, `a timer that runs while the page is hidden` in `useUI.test.js`, the visibility case in `viewport-guard.test.js`, `letting go early …` in `sound.test.js`.

### Checklist

- [x] `npm test` passes in `frontend/` — 1598 passing across 106 files
- [x] User-facing strings are in every locale pack — `node scripts/check-locales.mjs`: 14 locales, 1303 keys each, in sync (fourteen strings added)


DuarteSantos8:
Thanks, and the iOS sound analysis in #152 was right. This bundles four features in 1.6k lines; if you can split the #152 sound fix (silent-switch category, unlock, playback element, the setting) into its own PR it can go into v1.3.6, the rest after.

kurktchiev:
Done: #183 carries the #152 sound fix alone (silent-switch category via the setting, unlock from the tap, resume before every tone). This branch is rebased on top of it and shrinks to the per-kind rest-over sounds and the timer-bar work once #183 merges.

kurktchiev:
Rebased onto v1.3.7 (d7f0f0d). With #183 merged this is the remaining half only: per-kind rest defaults, the rest phase on the timer, and the finish-screen sound. Seven commits, suites green.

kurktchiev:
Pushed one more commit (`1246427`), found while training with this branch.

**The bar promises a move it never makes.** For the whole countdown it says `Next exercise · X`. Then the rest ends, the toast says "next set!", and the screen is still on the exercise you just finished — the only way on is *Next*. Same for a superset: `Next round · A`, countdown ends, nothing.

So every rest a completed set starts now carries a hand-over. When it runs out, or you tap Skip, `restFocusIdx`'s exercise becomes the current one — the exercise the bar has been naming all along, so the label and where you end up can't disagree. Like the hold chain it re-reads the store when it fires (a rest outlives the render that armed it), and it never overrules a move you made yourself during the break.

Where nothing is going to time the gap the move happens on the tick instead: rest timer set to Off, a next exercise that ramps through warm-up sets of its own, a backfilled session.

**On !92.** That decision — the marker no longer follows a finished exercise — still holds here: ticking the last set must not yank you away while you are still fixing a weight or rating effort, and it doesn't, the move waits for the rest to end. What I don't think !92 meant is the case where there is *no* rest either: before an incomplete warm-up row you got neither a countdown nor a way forward. Five cases that asserted "stays put" there now assert the move; happy to split that half out if you'd rather keep it strictly as-is.

**Second half, same symptom from the other side.** A `round` rest pointed at the first member of the superset even when that member had run out of sets. In a group whose members don't all have the same number of sets that sends you back to work you've already finished, while the set you actually owe is in the exercise you were just on. `restFocusIdx('round')` now takes the first member that still *has* work. Two cases in `supersetFlow.test.js`.

Frontend suite green: 106 files / 1546 tests.

kurktchiev:
One more on top (`db95f84`), same class, found while re-reading the tick.

The rest's hand-over was dropped entirely when the countdown had expired while the app was hidden. That rule is right for the hold chain — a rest that ran out in your pocket must never start a hold nobody watched, or it logs time you did not spend. It is wrong for the other half: lock the phone during the break, come back, and you are still on the exercise you finished before locking, which is exactly the case where you most want the screen to have moved.

So the rest now tells its hand-over which of the two it is: `onDone(forIdx, seenLive)` fires either way, `Workout.jsx` moves the screen on regardless and gates only the next hold on `seenLive`. `skipRest` passes `true` — you are looking at it. The `useUI.test.js` case that asserted "does not fire when the rest ran out while the app was hidden" now asserts it fires with `seenLive: false`, and there is a matching view case.

Frontend suite: 1547 passing across 106 files.

kurktchiev:
One more commit, from training with this branch: the count-in is now queued rather than ticked, and how loud it is is a setting. Section 6 in the description.

Short version: `if (left <= 3) beep(...)` ran on the timer's own 1 s interval, and that interval is throttled to a crawl in a backgrounded tab — which is where a phone spends a rest — so the count-in only happened when you were already looking at the screen. `countdown(enabled, secLeft)` queues the whole burst in Web Audio when the timer starts, inside the tap that finished the set, and audio queued that way plays on its own clock. Five ticks rather than three, on the hold timer as well as the rest timer. `hush()` is the other half: everything that ends a timer early already goes through `stopRest`/`stopWork`, so nothing can tick after you have moved on, and the first tick back on screen re-queues against the time really left (a locked phone freezes the audio clock with the page).

The volume half is smaller but it is the thing that was actually complained about: every tone ramped to a hard-coded 0.35 peak on a sine. Settings → Sound volume is `low` / `medium` / `loud` — `low` is exactly today's level, so nothing is taken away — defaulting to `loud`, including for existing settings. Happy to default it to `medium` instead if you would rather not change what existing users hear; it is one constant (`DEFAULT_VOLUME` in `lib/sound.js`).

Tests: 1577 passing across 106 files, `check-locales` in sync at 1303 keys.

kurktchiev:
Correction to the volume half, from testing it on the phone rather than reasoning about it: raising the peak from 0.35 to 1.0 did nothing audible. +9 dB on paper, no difference in the room.

The reason is the speaker, not the code. A phone driver has a couple of millimetres of excursion and sits behind a limiter, and a pure tone at 660–880 Hz is the worst case for it — all the energy in one low band it can barely move air in, so past a point more amplitude buys more limiting. What a small driver is efficient at, and what the ear is most sensitive to, is 2–4 kHz, which is exactly where a square wave's harmonics land (3f = 2 kHz for the countdown tick) without moving the note.

So the levels now differ in waveform as well as peak: `low` is the old pure sine at 0.35, `medium` is half scale with harmonics, `loud` is full scale with harmonics. Every frequency, pattern and duration in the file is untouched, so the set/round/block sounds still read as themselves — only the timbre hardens as you turn it up. `low` remains exactly what every build so far has sounded like, so nobody who likes the soft tone loses it.

Tests: 1578 passing across 106 files.

kurktchiev:
Second correction to the volume half, and it supersedes the square-wave comment above.

Making the louder levels a square did fix `low` vs `loud` — they are clearly different now. But `medium` and `loud` were the same square differing only by a multiplier (0.5 vs 1.0), and on the phone they came out the same as **each other**. Which is the same lesson twice: a limiter is free to erase any difference that is only a multiplier, so gain cannot be the thing that separates two levels.

The levels are now separated by where the energy sits, on two axes:

- `low` — sine, at the written pitch (unchanged from every build so far)
- `medium` — square, at the written pitch
- `loud` — square, **an octave up**, which moves the fundamental itself into 2–4 kHz where a phone driver is efficient and the ear is most sensitive, instead of relying on harmonics to get there

Modelled against a phone response and A-weighted: +8 dB low→medium, +5 dB medium→loud, holding even in the worst case where the limiter pins every level to the same peak. An octave means every pattern keeps its rhythm and its intervals — there is a test for exactly that — so the set/round/block sounds stay themselves, pitched up.

One negative result worth recording so nobody repeats it: I first tried a custom `PeriodicWave` with the 3rd/5th partials boosted and the fundamental pulled down. It measures *worse* than a plain square — normalised to the same peak its crest factor is higher, so it carries less energy. Reverted before it shipped.

Tests: 1581 passing across 106 files.

kurktchiev:
One more commit on the volume half, and with it I think the search is closed.

The ear integrates loudness over roughly 200 ms, so below that a tone twice as long is heard as about 3 dB louder for nothing — as much as the whole medium→loud step costs in spectrum. Each level now carries its own countdown-tick length: 0.1 / 0.15 / 0.2 s. The **countdown only**: its ticks are a second apart and have the room, while the rest-over patterns are 0.15 s apart and a 0.2 s tone there would smear three beeps into one and destroy the per-kind meaning that section 1 is about. There is a test pinning that.

I also went back and measured whether the octave was the right stopping point rather than assuming it, across all five notes this file plays: `x2.5` is +0.1 dB on `x2` and `x3` is already **-0.2 dB**, because a small driver rolls off above ~5 kHz as surely as it does below 1 kHz. So the octave is the plateau, not a compromise — worth recording so the next person does not reach for a bigger number.

That leaves all four levers at their limit: gain is eaten by the phone's limiter, a square is the most energy a waveform carries for a given peak, pitch is at its plateau, and duration is at the ear's integration limit. If a timer still is not audible enough for someone in a loud gym, the remaining move is leaving Web Audio entirely (a local notification's alert sound takes a different path through iOS), which is a different feature and not this PR.

Tests: 1583 passing across 106 files.

kurktchiev:
Last commit, and it is the one that explains why the previous four did not settle it.

With all four in-app levers at their limit — gain eaten by the limiter, a square carrying the most energy per peak, pitch at its measured plateau, duration at the ear's integration limit — a timer could still be too quiet. That is a strong hint the cause was never in the app, and it is not.

Under the `'playback'` audio session (section 2) the timer plays on the phone's **media** channel. iOS only points the hardware volume buttons at that channel while something is actually playing. `lib/sound.js` suspends the context a second after each burst, on purpose, so between two beeps the buttons were adjusting the **ringer** instead. A phone whose media volume happened to be low therefore had no reachable way to be turned up: pressing volume-up during a rest moved a slider that had nothing to do with the timer, and nothing the app did to its own output could ever have compensated.

`holdSession(on)` keeps the context running for the length of a rest or a hold (`startRest`/`startWork` hold it, `stopRest`/`stopWork` let go, and the idle timer skips suspending while held). The buttons then do the obvious thing for as long as the timer is running. The Sound volume row says so in its subtitle, because nothing in the browser can read the system volume, let alone set it.

Trade-off: under `'playback'` the phone's own music now stays paused for the whole rest rather than just across each beep. That is the same trade the setting already warns about, and it only applies to people who turned that switch on.

Tests: 1588 passing across 106 files.

kurktchiev:
Housekeeping, and an apology for the noise: my four comments above worked out the volume problem in public, and two of them correct the one before. They are superseded — please read only the PR description and this comment.

I have squashed those five commits into one (`d99ae4f`), so the history no longer shows me changing my mind three times about the same twenty lines. The tree is byte-identical to what was there before the force-push; 14 commits → 10.

**Where the volume work landed**, as one statement instead of a trail:

- Gain is not the lever. A phone speaker sits behind a limiter, so raising the peak from 0.35 to 1.0 (+9 dB on paper) was inaudible, and two levels differing only by a multiplier came out identical to each other.
- The levels are separated by spectrum instead, none of them by gain alone: `low` = sine at the written pitch (what every build so far has sounded like), `medium` = square at the written pitch, `loud` = square an octave up, which moves the fundamental itself into the 2–4 kHz band a small driver is efficient in.
- The octave is the measured plateau, not a compromise: ×2.5 is +0.1 dB and ×3 is −0.2 dB, because small drivers roll off above 5 kHz too.
- Each level holds its countdown tick slightly longer (0.1 / 0.15 / 0.2 s) — about 3 dB under the ear's ~200 ms integration window. Countdown only; the rest-over patterns are 0.15 s apart and would smear.
- And the reason none of that was sufficient alone: Web Audio plays on the media channel, and iOS only aims the hardware volume buttons there while audio is actually playing. Suspending between beeps left those buttons on the ringer, so a phone with low media volume could not be turned up at all. `holdSession()` keeps the context running for the length of a timer.

Two negative results, recorded so nobody repeats them: a `PeriodicWave` with boosted 3rd/5th partials measures **worse** than a plain square (higher crest factor, less energy at the same peak), and raising gain alone does nothing on a phone at any level.

On size — you asked back in September to split this and I did (#183), but it has since grown to +1917 rather than shrunk, which is not what you asked for. The offer in the description stands: say the word and sections 5 and 6 become their own PRs on top of whatever you merge first. Confirmed working on the reporter's phone; CI green.

kurktchiev:
Three more commits, from a real session on the phone with this branch (18 Sept):

- `d2739bd` — the screen only ever moves forward. `nextUnfinishedUnit` wraps (it answers whether a rest is owed), and as a destination that sent the card back to a skipped warm-up block after every later exercise. Hand-over and on-tick move now use `nextUnitAhead`; the bar still names the wrapped exercise. A rest owed by a re-check moves nothing.
- `aa30c37` — `holdSession(false)` really sleeps the context: the queued countdown's deadline made `sleepAfter(0)` a no-op, so a skipped rest kept the session (and the phone's silence) alive to the original end.
- `a2bff62` — a hidden page changes nothing on screen. The held session keeps an iOS page running behind the lock screen, so the ticks completed the rest unseen: double "rest over" alerts (local + push) and the fixed bars left mid-page. Hidden ticks now wait for the return tick; the local notification carries the push's tag; the viewport guard realigns on return.

Description updated (section 5 addendum, new section 7). 1598 tests across 106 files pass; the deployed fork runs this tree.

kurktchiev:
Ordering note: this one should land after #251 — https://github.com/DuarteSantos8/openGym/pull/251#issuecomment-5766864320

Both PRs give `startWork`'s `onDone` callback a second argument, and they are different ones (this PR: the held entry's current index, so a hold is never written onto an exercise that moved; #251: a flag for a hold a rest displaced). A trial merge conflicts in four hunks across `store/useUI.js` and `views/Workout.jsx`, so whichever goes second needs a design merge rather than a rebase.

#251 is 4 commits over 6 files and this is 13 over 36, so it is much cheaper that way round. Once #251 lands I will rebase this on top, fold both arguments into a single object, and settle one behaviour question the combination raises: the self-running timed exercise added here chains hold → rest → next hold on a guard that checks the *next* set is undone, and #251 introduces a state that guard has never seen — a displaced hold whose *current* set is left unticked. Whether the chain should advance there or wait for a tap is a real decision, and it can only be made with both in the tree.

Rebased onto v1.3.8 and green in the meantime.

Issue #158, opened 2026-09-11
[Feature Request] Sequential Routine Rotation (A → B → C → A) with Dynamic Rerouting


DuarteSantos8:
> Migrated from GitLab issue [#79](https://gitlab.com/DuarteSantos8/opengym/-/issues/79), opened by **Giulio Leuci** (`@leuci.giulio` on GitLab) on 2026-09-09. openGym moved back to GitHub on 2026-09-11; the GitLab thread is closed and points here.

**Labels**: `enhancement`, `scheduling`, `rotation`, `data-model`

#### 1. Current State

openGym's weekly planner (`frontend/src/store/useStore.js`) is strictly **fixed-week**:

- `S.week` maps each weekday to a **fixed** list of routine ids (`S.week[1] = ['r1','r2']`), resolved by `effectiveRoutineIds(S, iso)` in `frontend/src/lib/history.js` (precedence: `S.dayPlan[iso]` override → `S.week[getDay()]` → `[]` = rest).
- `S.dayPlan` supports one-off per-date overrides (scalar routine id or the `'rest'` sentinel), but reverts to the fixed week the next day.
- `frontend/src/lib/starter.js` documents the ceiling explicitly: _"Fixed weeks only: every plan repeats the same seven days, which is all the weekly plan model can represent."_ The starter catalog models A/B alternation as four **distinct routines pinned to specific weekdays**, not as a sequence.
- Programs that do **not** fit a weekly cadence (PPL with drifting training days, alternating A/B full-body, rolling 4-day splits) are impossible to represent: users must hand-pick a routine every single session.

#### 2. Motivation

- **Cycle-Based Training**: Rotation (A → B → C → A → B…) is the canonical model for A/B alternation (PPL, Upper/Lower) and volume/intensity cycling. The fixed-week model requires duplicating each routine onto separate weekdays, which breaks as soon as a day is skipped or a session is deferred.
- **Free-Running Cadence**: Rotation decouples _what you do next_ from _which weekday it is_. Training Monday/Wednesday/Friday — or any other cadence — simply consumes the next routine in the sequence.
- **Daily Autonomy with Guidance**: The system _suggests_ the next routine, the user may override at any moment, and the rotation re-routes itself from whichever routine was actually executed. The suggestion guides, never blocks.
- **Zero Learning Cost**: Rotation is an **additional** scheduling mode beside the fixed week. A rotation cycle that is never activated changes nothing for existing users.

#### 3. Feature to Implement

[...]

PR #167, opened 2026-09-11
feat: an ordered session queue beside the fixed week (S.queue) — free-running order, pins, progress row (#158)

kurktchiev:
Relates to #158 (sequential rotation, free-running cadence). This is the "next session is the next undone one, whatever the weekday" half of that request, built as an additive scheduling mode beside the fixed week, with a small state field an external planner (or a future in-app rotation editor) can write. Four commits, rebased onto current `main`. Off unless `S.queue` is set: an existing profile sees no change.

### The model

```js
S.queue = null | {
  ids:      string[]      // routine ids in order = the week's sessions
  since:    number        // ms epoch of the apply; a workout STARTED at/after it is this week's
  startsOn: 'YYYY-MM-DD'  // local date the queue becomes active
  label:    string        // progress-row title, e.g. 'Week 3'
}
```

- **Order, not weekdays.** The first undone session is today's session, every day, until it is done. Nothing is attached to a weekday. Miss Monday, train Tuesday: Tuesday shows Monday's session, and the rest shift with it. No week is left half-done because the calendar moved on.
- **Done is decided by the log, in one place** (`lib/queue.js`, `DONE RULE`): a session counts once a finished workout on that routine exists after the queue was applied, or is dated inside the week and named for it (merged sessions read `'A + B'`). The api's reminder helper carries a hand-copied twin of the rule with the same tests, so push reminders and the Home screen never disagree.
- **Your own routines ride along.** A weekday routine of your own combines with the day's queue session, as combined days already do; `'rest'` and routine overrides still win for their date.
- **Pins.** Tap a day in the calendar, pick a queue session, and the existing per-date override (`S.dayPlan[iso] = <routine id>`) becomes a pin: that session leads on that day, is skipped by the floating rule on other days, and once done its pin is fulfilled and reads as no override. A pin on a past date is stale and the session floats again. Nothing new is stored.
- **The app never writes `S.queue`.** It is written through the existing state endpoint by whatever plans the week: today a planner talking to the API, tomorrow the in-app rotation editor #158 describes could refill it from a sequence when the week completes. `DEF.queue = null`.

### What you see

- Home replaces the seven weekday dots with a progress row while a queue is set: one chip per session in order, the next one lit, done ones ticked, a pinned one wearing its day, and a status line ("Next: Pull, today", "Next week starts Mon 14 Sep", "Week complete, ask the coach").
- The streak card's "done / planned this week" counts the queue's sessions and your own weekdays together (`weekTally`), so following the app's own suggestion all week ends at n / n.
- The day sheet lists the remaining queue sessions first, under "Coach week"; picking one pins it. Its header names the day's plan including the queue session.
- Push reminders (api) follow the same rule. The read-only MCP bridge's `get_week_plan` reports the next seven days with how each was decided (`coach`, `pinned`, `override`, `rest_override`, `weekday`, `rest`) and the queue's state.

On wording: the UI strings say "coach" ("Coach week", "Week complete, ask the coach") because the thing that writes a week is, to the user, a coach — a person, a program, an agent. Happy to change the noun if you prefer something else; it is seven locale keys.

### Where

- `frontend/src/lib/queue.js` (+ `queue.test.js`): the tolerant reader, DONE RULE, `queueNext`, `queueView`, `weekTally`, `pinState`, `queueLiveOn`.
- `frontend/src/lib/history.js`: one hook in `effectiveRoutineIds(S, iso, today)` — every existing reader (Home, TabBar, start sheet, calendar, next training day, mobile reminders) gets the behaviour through it.
- `frontend/src/components/QueueRow.jsx`, `views/Home.jsx`, `sheets.jsx` (day sheet, calendar dot), `views/Workout.jsx` (one line).
- `api/queue.js` (+ tests) and the reminder tick in `api/server.js`; `api/openapi.yaml` State schema gains `queue`; `api/Dockerfile` copies the module, and `api/test/dockerfile-copies.test.js` guards that list against the next module.
- `mcp/src/tools.js` `get_week_plan` (+ tests) — reads the frontend's lib directly, no third copy.
- Seven new strings in all 14 locales.

### How it relates to #158

#158 asks for two things: a free-running order instead of weekdays, and an in-app sequence editor with rerouting from the last logged routine. This PR is the first: the data model, the rule, the Home/calendar/reminder surfaces, pins, and the "done" semantics that make rerouting well-defined (a session is done when it is logged, wherever it was logged). A rotation editor on top of it is a writer of `S.queue`: when the queue completes, refill it from the sequence. I did not build that editor here; the issue's author has a working prototype and a design for it, and this keeps the two pieces separable.

### Checklist

- [x] `npm test` passes in `frontend/` — 1402 passing across 96 files
- [x] `npm test` passes in `api/` — 174 passing
- [x] `npm test` passes in `mcp/` — 66 passing; `node scripts/check-node-loadable.mjs` passes
- [x] User-facing strings are in every locale pack — `node scripts/check-locales.mjs`: 14 locales, 1272 keys each, in sync (seven strings added)


DuarteSantos8:
Thanks — this is the free-running half of #158 done properly. Own review after v1.3.6; I'll comment on the model in the PR.

kurktchiev:
Rebased onto v1.3.8. Clean merge state again — api 213/213, frontend 1658/1658, mcp 66/66.

The rebase carried upstream's `?.` hardening of `effectiveRoutineId` into the copy this PR moves out to `api/queue.js`, and picked up the new strings in the fourteen locale packs.

You said on 2026-09-12 you'd comment on the model after v1.3.6. v1.3.7 and v1.3.8 have shipped since, and each one has needed a rebase here — this is 38 files, so the drift is real work rather than a formality.

Happy to keep it current, and equally happy for it to wait. What would help is knowing which: if the model needs changing, tell me what and I'll rework it; if it should sit until the v1.4.0 database lands, I'll mark it draft and stop rebasing. Either is fine — I'd just rather not keep re-landing a design you may not want.

Issue #153, opened 2026-09-11
[Feature Request] Advanced Intensity Prescriptions: Flexible Load Specification (Absolute vs %1RM vs Auto) and Target Perceived Exertion (RPE/RIR)


DuarteSantos8:
> Migrated from GitLab issue [#74](https://gitlab.com/DuarteSantos8/opengym/-/issues/74), opened by **Giulio Leuci** (`@leuci.giulio` on GitLab) on 2026-09-09. openGym moved back to GitHub on 2026-09-11; the GitLab thread is closed and points here.

~idea 

**Labels**: `enhancement`, `training-engine`, `data-model`, `autoregulation`

#### 1. Current State

In openGym, routine exercise prescriptions are strictly static and one-dimensional:

- In routine configurations (`routine.ex[]` / `cfg`), training load is stored only as a single fixed number in `cfg.weight` (e.g. `cfg.weight = 80`). The helper `exLine(cfg, unit)` in `frontend/src/lib/history.js` formats this as a static string (e.g. `3 × 10 · 80 kg`). When launching a workout session in `frontend/src/views/Workout.jsx`, `buildSets()` simply copies this value directly into the working set rows (`set.w = cfg.weight`).
- While openGym possesses a subjective exertion logging mechanism in `frontend/src/lib/effort.js` (`s.rir` / `s.rpe`), it is **strictly retrospective**: athletes can log how hard a completed set felt, but coaches and lifters cannot _prescribe_ a target intensity level (such as `@ RPE 8` or `@ 2 RIR`) when creating or editing a routine.

#### 2. Motivation

- **Scalability with Athlete Progression**: In percentage-based training (Powerlifting, DUP, Peaking blocks), programs prescribe loads as percentages of 1RM (e.g. 5×5 @ 75% 1RM). Without native %1RM support, users must manually calculate and edit plate weights for every exercise whenever their strength changes.
- **Graceful Fallback**: Providing an "Auto" mode allows percentage-based programming while ensuring that if an athlete does not yet have an estimated 1RM for an exercise, the routine falls back to a baseline absolute load without breaking.
[...]

PR #168, opened 2026-09-11
Percentage (wave) progression + SetBlock refactor

giulioleuci:
## Summary

This PR adds percentage-based ("wave", 5/3/1-style) progression to openGym, and then refactors its internal data model before it ships. It's two pieces of work stacked together:

**1. Percentage/wave programming** (already reviewed as [MR !136](https://gitlab.com/DuarteSantos8/opengym/-/merge_requests/136) on the GitLab canonical repo — included here so this branch applies cleanly on GitHub):
- A new `wave` progression policy: multi-stage percentage cycles off a training max (linear, Greyskull LP, double progression, time-based rules already existed — this adds the 5/3/1-style one)
- Per-row prescription (each set in a stage can have its own weight/reps), row-by-row grading of a logged session
- A wave editor in the exercise progression sheet, wave-aware workout display, PDF plan export support
- Wave settings travel through plan sharing, unit conversion, and the AI Coach's plan payload and change-detection fingerprint (both frontend and API)

**2. `SetBlock` refactor** (this session's work, on top of the above):
- Replaces the wave's flat `stage.sets[]` array with `stage.blocks[]` — a `SetBlock` (stable id, `sets`/`reps`/`pct`/`type`, and a `role: 'anchor'|'required'` derived at read time, never persisted) — so a stage's sets are addressable, named blocks instead of an anonymous list
- `prescribeWave`/`readSession` now grade a wave session by its explicit **anchor** row instead of inferring the "heaviest logged weight" — more correct, and it fixed a real bug in the process (see below)
- Every row a wave produces carries its `blockId`/`stageId`/`role` from prescription through to saved workout history
- The Coach's plan-change fingerprint now hashes the *normalized* wave shape instead of the raw stored field, in both the frontend and the API (kept in parity by a dedicated test), so editing a wave's percentages or training max is correctly detected as a plan change
- **Bug fix included:** the anchor-based grading initially had a regression where a session with the anchor set logged-but-not-ticked would be graded as `0`, causing the progression to jump to a deload stage instead of repeating the current one. Caught in review before merge, fixed, and pinned with a test that drives the real prescribe → log → finish → prescribe loop (not just isolated units)

## Test plan

- [x] `cd frontend && npm test` — full suite passing (93 files / 1397 tests)
- [x] `cd api && npm test` — full suite passing (157/157)
- [x] Manual pass on the wave editor and active-workout UI recommended before release (component-level UI wasn't smoke-tested against a live browser for the block-role header and the editor's new controls — covered by unit tests on the underlying data, not by an E2E click-through)

https://claude.ai/code/session_01K8rxrBQzG2Hfx65fxZxPex


DuarteSantos8:
Thanks. This gets its own review after v1.3.6 — it refactors the set model, and I want to read it with #181 and #153/#154 in mind rather than squeeze it into a bug-fix release.

giulioleuci:
@DuarteSantos8 I plan to merge #179, #181, and #168, reimagining them within a more generic, composable, and reusable common engine. The new issue is #186. I'll implement it in the next few days if you like the proposal. I hope this will also make your job easier.

PR #171, opened 2026-09-11
i18n: add de-CH as a derived locale

mkoester:
Adds Swiss German (`de-CH`) as a selectable language. It ships **no locale pack of its own**: `de-CH` is derived from `de` at load time by a single substitution — Swiss Standard German writes `ss` for every `ß` — applied to the locale pack and, when they exist, to the instruction and exercise-name packs too.

## Why derived, rather than a `de-CH.js`

- A hand-maintained pack would be **99.4% identical** to `de.js`: 8 of 1265 values differ (16, if the companion change below lands). For contrast, `pt` and `pt-BR` differ in **42.2%** of values, which is what earns *that* pair two files.
- `check-locales.mjs` enforces an identical key set across locales, so a second German pack would oblige every future German string to be written twice, forever — and almost all of those pairs would be identical.
- The transform is exact in **one direction only**. Going back needs vowel length (*Maße* and *Masse* both collapse to *Masse*), so `de` is always the base and never the derivative.

## The part that isn't about spelling

`DATE_LOCALES` drives number formatting, and the two differ there:

```
de-DE   1.234,5
de-CH   1'234.5
```

A Swiss user logging 102.5 kg currently sees `102,5 kg`. That's the practical reason this seemed worth doing — the spelling is arguably the lesser half.

## Why I'm working on this

I'd like to contribute **German exercise instructions**, which CONTRIBUTING lists as a welcome contribution. One wrinkle worth flagging: the exercises dataset ships `instruction_steps` for `en, es, fr, hi, it, ko, pl, ru, tr, zh` but **not** `de` — I checked all 1,324 entries — so German has to be curated rather than generated from the dataset. That work is under way on a separate branch and isn't part of this PR.

Getting there meant settling German orthography first, and `de-CH` falls out of it for free. Once the instructions land, `_setLangState` tests `INSTR_LANGS.includes(baseLang(lang))`, so **de-CH inherits them automatically** — no second list entry to remember.

## A companion change, deliberately kept separate

There's a related but independent branch, `fix/de-locale-sharp-s`, which normalises `de.js` to standard `ß`. I've kept it out of this PR because it's a judgement call and this one isn't.

The short version: the German locale currently spells the same word both ways — `'Glutes': 'Gesäß'` beside `'glutes': 'Gesäss'`, `ausschließen` beside `abschliessen`, `großen` beside `Gross` — and `DATE_LOCALES` already maps `de` to `de-DE`. That reads as drift rather than a deliberate choice, but it's your call and I'd rather ask than assume. I'll open it separately, and I'm happy to drop it if you'd prefer the base locale stayed as it is.

**This PR works either way**: `de-CH` renders `ss` whichever way `de.js` is spelled.

## Tests

New `src/lib/i18n-core.test.js` — 13 tests covering `baseLang`, `derivePack` (flat packs, nested instruction arrays, non-derived passthrough, `null`), the whole-locale derivation, and the `de-CH` number formatting above. The whole-locale test carries a control asserting `de.js` still contains `ß`, so it can't pass vacuously.

Four existing test files assert one locale pack per `LANGS` entry. A derived locale is the first exception to that, so they now compute the count as `LANGS − en − DERIVED_LOCALES`; `i18n-core.test.js` asserts the other half, that a derived locale has no pack file of its own.

`npm test` on Node 22: **94 files / 1343 tests** green.


DuarteSantos8:
Reviewed. A derived locale rather than a second pack is the right call; lined up for v1.3.6 after running the locale guard tests.

PR #174, opened 2026-09-11
feat(settings): a switch to skip the weigh-in when a workout starts (#137)

### LINKED ISSUE #137: do not force weight check in at start
> Migrated from GitLab issue [#53](https://gitlab.com/DuarteSantos8/opengym/-/issues/53), opened by **Boris Kurktchiev** (`@kurktchiev` on GitLab) on 2026-09-05. openGym moved back to GitHub on 2026-09-11; the GitLab thread is closed and points here.

Currently I cannot turn off the weight check in at the start of any workout. It is fine if I am in a cutting period but more distracting step otherwise. Looking for an option under the settings to drop the weight check on start basically.

kurktchiev:
Closes #137.

**What**

A switch in Settings → During a workout: **Weigh in before workouts** (on by default). Off, tapping Start begins the session straight away, with no body weight on it — the same thing "Start without weighing in" did, but without the sheet. Weight can still be logged by hand from the Home and Stats "Log" buttons, and the Home body-weight footer no longer promises "it's also asked before every workout" while the switch is off.

**How**

- `S.weighIn` (DEF `true`; read as `!== false`, so a profile written before the key existed keeps asking, same pattern as `checkIn`).
- `startFlow` in sheets.jsx is the single entry point for every Start button (Home today/queue rows, the workout picker, freestyle); it now checks the setting before opening `bwSheet`. "Log a past workout" never asked and is unchanged.
- Three new strings in all 14 locales; pt-BR carries explicit overrides so the inherited pt-PT set and its fingerprint are unchanged (test count 631 → 634).
- `sheets.weighin.test.jsx`: on / legacy-profile / off, through the real store and sheet stack.

**Verified**

`npm test` in `frontend/`: 94 files, 1333 tests pass, plus `scripts/check-locales.mjs` (14 locales in sync) and `npm run build`.


DuarteSantos8:
Reviewed, CI green, lined up for v1.3.6. Thanks.

PR #177, opened 2026-09-12
Add distance logging mode (metres stored; feet for imperial)

baron-de-ropp:
## Summary

Adds a **distance** logging mode for carries and loaded walks — time cap + farthest distance per set — with imperial profiles entering and displaying **feet** while storage stays metres (same labels-only rule as weight).

This PR **includes** the former feet-entry follow-up (#178); that PR is closed as superseded.

### Logging / storage
- Set shape: `{ sec, m }` (time cap + metres covered)
- Metres on disk; lb profiles enter/show feet via `metresToDisplay` / `displayToMetres` / `distanceUnitLabel`
- Wired through config sheet, in-session Workout columns, Stats farthest-distance, CSV import, plan share

### Progression
- **Add distance** policy (+10 m default), hold on miss, deload after repeated misses (`DELOAD_AFTER.distance = 3`)

### Modes registry
- `distance` registered in `workout-model` `MODES` and `METRIC_MODES`
- Inference prefers `.m` over bare `.sec` so carries are not read as timed holds
- Exercise config always offers Reps / Time / **Distance** (non-cardio)

### Locales
- All **14** packs updated (including `hu` / `th`)
- Progression why-strings and `Distance ({0})` / `Step ({0})` for unit labels

### Tests
```bash
cd frontend && npx vitest run \
  src/lib/history.test.js \
  src/lib/progression.test.js \
  src/lib/import-distance.test.js \
  src/lib/workout-model.test.js \
  src/lib/pt-br-locale.test.js
```

## Test plan
- [ ] Set an exercise to **Distance** from Reps/Time (mode control always shows Distance)
- [ ] Log a session: time cap + distance; finish once — workout appears **once** in History
- [ ] lb profile: entry shows **ft**, stored value still metres; switch unit and confirm display flips
- [ ] Stats: farthest distance uses profile unit
- [ ] Progression: hit target → distance steps up; miss repeatedly → deload
- [ ] Commands above pass; `node frontend/scripts/check-locales.mjs` → 14 locales in sync


DuarteSantos8:
A distance mode for carries makes sense. Two things before it can go in: (1) it lands after v1.3.6, and #178 (feet entry) should be squashed into it so the mode never ships half-converted; (2) the new strings need to be in all 14 locale packs (`node frontend/scripts/check-locales.mjs` and the pt-BR guard test tell you when it's complete).

DFW720:
Thanks for this — distance + time cap is useful for loaded carries and rucks.

One gap for pack / ruck training: the proposed set shape is `{ sec, m }` with no load. For a ruck I’d want to log **weight (pack load) + distance + time** on the same set (e.g. 45 lb for 3 miles in 48 min), and eventually progress on weight and/or distance.

Would you consider extending distance mode so sets can optionally carry `w` the way timed loaded carries already can? Happy to clarify the UX if helpful.

PR #178, opened 2026-09-12
Enter and edit distance in feet for imperial profiles

baron-de-ropp:
## What

Imperial (lb) profiles now enter and edit distance in feet on every entry surface, while the stored `.m` stays metres — the same labels-only unit rule weight already follows.

## Why

PR #177 added the `distance` logging mode and converts the *display* to feet for lb profiles, but the *entry* surfaces (exercise config Distance stepper, progression Step stepper, in-session logging column) still asked for and showed raw metres with labels like `Distance (m)`. A user composing a 400 ft carry unintentionally stored 400 metres.

## Changes

- `history.js`: three pure helpers — `metresToDisplay`, `displayToMetres`, `distanceUnitLabel` — with unit tests for the round-trip and kg passthrough.
- `sheets.jsx` (config sheet): Distance and progression Step steppers convert feet ↔ metres at the edit boundary; labels show `ft` for lb, `m` for kg.
- `Workout.jsx` (in-session table): the distance column shows/steps in feet for lb, converting on both read and write so `s.m` stays metres.

No i18n keys were added, renamed, or removed — the `ft` labels are computed at runtime — so no locale files or pt-BR inheritance fingerprint changed.

## Notes

- Built on top of PR #177 (this branch contains that commit). Once #177 lands on `main`, this diff narrows automatically to the feet-entry change.
- Distance is stored in metres so the data is unit-neutral; switching units changes what is entered/shown, never what is stored.

## Tests

`npx vitest run src/lib/history.test.js src/lib/progression.test.js src/lib/import-distance.test.js src/lib/plan-share.test.js src/lib/pt-br-locale.test.js` → 272 passing; plus `src/views/Workout.stepper.test.jsx` (4 passing).


DuarteSantos8:
Reviewed together with #177 — see there.

baron-de-ropp:
Superseded by #177 — feet entry is folded into the polished distance-mode PR so imperial users never get metres-only entry.

PR #181, opened 2026-09-12
feat: add triple progression

### LINKED ISSUE #179: Triple progression: reps → sets → load
### What problem does this solve?

Extend `frontend/src/lib/progression.js` (and friends) with a new `triple`
policy, separate from `double`.

Triple progression advances three variables in order: **reps → sets → load**.
The term is established training vernacular, not formal science, but maps
onto the three distinct overload levers literature recognizes. `double` keeps
its current behavior unchanged (reps to the top, then load) — no set ranges,
no rename.

### Configuration

`triple` uses `setsMin`/`setsMax`, `repsMin`/`repsMax` (`reps` still stores
`repsMax`), shared `inc`, and `deloadFactor` with the existing Epley/fallback
behavior. Validate/normalize both ranges: positive integers with
`setsMin <= setsMax`; `repsMin <= repsMax` honoring `repStep(cfg)`. An
incomplete `triple` config must not yield ambiguous targets — the UI requires
all four limits.

### Cycle

Example: 3–5 sets, 8–12 reps.

```
8, 8, 8 → … → 12, 12, 12 → 12, 12, 12, 8 → … → 12, 12, 12, 12
→ 12, 12, 12, 12, 8 → … → 12, 12, 12, 12, 12 → load + inc; 8, 8, 8
```

Base sets (`setsMin`) advance together from `repsMin` to `repsMax`; each
added set climbs alone from `repsMin` to `repsMax` before another is added;
mature sets stay at `repsMax`. For a completed session:

1. Any set under target → keep same load and same `rows` (no target cut, unlike `double`).
2. Base group active, all on target → step all by `repStep(cfg)`, capped at `repsMax`.
3. Base at `repsMax` or the active added set on target but under `repsMax` → step only the active added set.
4. Active set at `repsMax` and sets < `setsMax` → add a set at `repsMin`.
5. All `setsMax` sets at `repsMax` → load + `inc`, prescribe exactly `setsMin` × `repsMin`.

Return `rows` with a per-work-set target, plus `weight`/`sets` when useful;
`applyPrescription()` applies without flattening mature sets.

### Misses, stalls, deload

A session fails per the common engine rule (missing/missed/under-target/unchecked set). Under `triple` a single failure or an under-target improvement does **not** regress `rows` — repeat the prescription. But any rep gain in the active series/group at the same load is partial progress and breaks the stall streak; compare against the base group's minimum while it's active, else the most recent added set. A load change always breaks the streak.

After **three real stalls**: existing deload — Epley load at `repsMin`, else −10%; restart at `setsMin × repsMin`. `stallCount()` must treat `triple` as partial-progress but cannot blindly reuse `double`'s `last.low`.

### Scope

- Add `triple` to `POLICIES`, `POLICIES_FOR.reps`, names, descriptions, i18n, MCP labels.
- Extend `ProgressionFields` with set/reps limits only for `triple`; `double` UI/semantics unchanged.
- Normalize/validate `triple` config where exercise progression config is already handled.
- Add `policy === 'triple'` to `nextPrescription()`; derive phase only from last prescription + finished session — no persisted counters.
- Extend session summary and `stallCount()` with the active series/group value to tell partial progress from stall.
- `applyPrescription()` may only shrink sets while creating a new session; never remove a set already logged in an in-progress one.
- Bodyweight: no external load, so `triple` stays unselectable there until an explicit harder-variation/weighted semantics exists.

### Tests (`frontend/src/lib/progression.test.js`)

1. Base group climbs uniformly `repsMin` → `repsMax`.
2. Single added set at `repsMin`, previous sets stay at `repsMax`.
3. Only the added set advances.
4. `setsMax × repsMax` → load up, `setsMin × repsMin`.
5. Under-target session: same `rows`, same load.
6. Partial progress under target: no deload after three sessions (active value improves).
7. Three real stalls: deload + sets/reps reset.
8. No regression for existing `linear` and `double` configs.

### Dependencies

- [x] This doesn't require adding a new dependency

giulioleuci:
## Summary

- add triple progression (reps → sets → load), including stalls and deloads
- add configuration UI, bodyweight safeguards, and MCP label coverage
- preserve existing double progression behavior

## Verification

- npm --prefix frontend test (1,353 passing)
- npm --prefix mcp test (59 passing)
- npm --prefix frontend run build

Closes #179.


DuarteSantos8:
Thanks. This gets a proper review after the v1.3.6 release, together with #168 (both change progression.js and the exercise config sheet).

giulioleuci:
@DuarteSantos8 I plan to merge #179, #181, and #168, reimagining them within a more generic, composable, and reusable common engine. The new issue is #186. I'll implement it in the next few days if you like the proposal. I hope this will also make your job easier.

PR #182, opened 2026-09-12
fix image ref to github

andi242:
change the image reference from gitlab back to github.

it might just work when gitlab is also building on the mirror, but since GH is the official source now....


DuarteSantos8:
Thanks — merging for v1.3.6.

PR #183, opened 2026-09-12
fix(sound): iOS timer sounds — resume before every tone, unlock from the tap, "play on silent" setting (#152)

kurktchiev:
The #152 sound fix on its own, split out of #165 as requested. `lib/sound.js`: every tone resumes the context first (a lock or app switch leaves it suspended), the context sleeps a second after the last tone, `unlock()` runs from the taps that lead to a timer, and Settings → Sounds gains "Play sounds when the phone is on silent" (`S.soundOnSilent`, off by default, iOS 17+ only), which sets the page's audio-session type to `playback`. Two strings in all 14 packs.

Tests: `lib/sound.test.js` (fake AudioContext and audioSession), `views/Settings.sound.test.jsx`. Frontend suite 101 files / 1432 tests, build clean.

#165 is rebased on top of this branch; once this merges it shrinks to the per-kind rest-over sounds and the timer-bar work.


DuarteSantos8:
Merged, thank you for the quick split. v1.3.6 was already tagged and building when this landed, so it ships as v1.3.7 right behind it (same day) — that will be the APK the site promotes. #165 can now be rebased to its remaining half.

Issue #155, opened 2026-09-11
[Feature Request] Multi-Formula 1RM Estimation with Reliability-Weighted Average


DuarteSantos8:
> Migrated from GitLab issue [#76](https://gitlab.com/DuarteSantos8/opengym/-/issues/76), opened by **Giulio Leuci** (`@leuci.giulio` on GitLab) on 2026-09-09. openGym moved back to GitHub on 2026-09-11; the GitLab thread is closed and points here.

**Labels**: `enhancement`, `training-engine`, `algorithm`

#### 1. Current Statea

In `frontend/src/lib/onerm.js`, openGym currently implements only three 1RM equations: Epley, Brzycki, and Lombardi.

- Epley is hardcoded as the default formula.
- Sets above `REP_CAP = 12` are rejected.
- Each formula has recognized structural biases: Epley progressively overestimates 1RM at moderate-to-high repetitions (\> 6 reps), Brzycki diverges sharply near 10–12 reps, and Lombardi is overly conservative at low repetitions.

#### 2. Motivation

Estimated 1RM is a critical metric in openGym: it directly powers percentage-based load resolution (%1RM), the in-session load calculator, and PR tracking. Relying on a single equation introduces systemic distortion. A multi-formula composite average weighted by empirical reliability and penalized for high reps creates a stable, smooth, and robust estimate across all rep ranges up to 15 reps.

#### 3. Feature to Implement

1. **Extend Formula Suite**:
   - Add 4 peer-reviewed submaximal equations: O'Conner, Mayhew, Wathan, and Lander.
2. **Reliability-Weighted Average (`weightedEstimate`)**:
   - At r = 1: Return the exact lifted weight (1RM = w).
   - For r \> 1: Compute estimates across all 7 equations. Weight each formula by its empirical reliability factor a_i and apply a logarithmic rep penalty factor w_i = \\frac{1}{\\log\_{10}(r) + 0.1} \\cdot a_i.
   - Set the default formula in openGym to `'weighted'`.
   - Extend the safe calculation cap to `WEIGHTED_REP_CAP = 15`.
3. **Preserve Legacy Signatures**:
   - Preserve exact backwards compatibility for explicit calls (e.g. `estimate1RM(w, r, 'epley')`).

#### 4. Technical Design & Code Changes

```javascript
// In frontend/src/lib/onerm.js






export const FORMULAS = {
  epley:    (w, r) => w * (1 + r / 30),
[...]

PR #195, opened 2026-09-13
1RM: seven-formula estimation suite with RIR handling, weighting, and accuracy index (#155)

giulioleuci:
## Summary

Implements **issue #155** (milestone **v1.3.12**): extends the estimated-1RM module with a seven-formula suite, RIR-aware estimation, a formula weight matrix, a confidence index — and routes the fatigue model through the same estimate.

### `frontend/src/lib/onerm.js`

**Formula suite.** Keeps Epley, Brzycki and Lombardi; adds O'Connor, Mayhew, Wathan and Lander in a single `FORMULAS` map.

**RIR handling.** The `estimate1RM(w, r, formula, rir)` signature gains an optional `rir` parameter. Effective reps (`reps + rir`) are read against a Mike Tuchscherer / RTS `%1RM` map, and the RIR-based estimate `weight / (%1RM / 100)` participates in the weighted ensemble.

**Weight matrix.** The flat rep-based cutoff is replaced by a two-tier per-formula weighting: an absolute reliability factor `a_i` multiplied by a variable rep attenuation `v_i(r)` (`weight_i = a_i * v_i(r)`). E.g. Brzycki decays rapidly past 8 reps, Epley overestimates past 6, the RIR map loses accuracy past 10.

**Weighted estimate.** A new `weighted` formula blends all seven formulas (plus the RIR map when RIR is present), capped at the new `WEIGHTED_REP_CAP = 15`; individual formulas keep `REP_CAP = 12`. `weightedEstimate()` is exported unrounded for internal consumers.

**Accuracy index.** `calculate1RMAccuracy(reps, rir, formula)` returns a 0–1 confidence index combining:
1. rep-count decay (highest at 1 rep, dropping toward `WEIGHTED_REP_CAP`),
2. RIR subjectivity (~8% penalty, only when `rir > 0` — RIR 0 is an objective to-failure set),
3. inter-formula coefficient-of-variation spread (for the `weighted` formula).

**Backward compatibility.** Existing callers are unaffected; a single rep still returns the measurement unchanged, and `r === 1, rir === 0` returns the load exactly.

### `frontend/src/lib/recovery.js`

The fatigue model's session-local 1RM anchor now uses the **onerm weighted mean** instead of a hand-rolled Epley clamp (removed the duplicated formula and rep cap). The unrounded blend keeps kg↔lb equivalence and the causal-reference math exact. Sets beyond `WEIGHTED_REP_CAP` form no anchor and register as unweighted raw stimulus — honest rather than scaled by a fantasy 1RM.

### Tests

`frontend/src/lib/onerm.test.js` gains coverage for the new formulas, the weighted estimate, RIR cases, and the accuracy index; `recovery.test.js` fixtures now derive from the weighted anchor instead of hardcoded Epley intensity. Full frontend suite: **1492 tests passing**.


giulioleuci:
Implement #155

PR #231, opened 2026-09-16
fix: invert PR and progression for assisted exercises

### LINKED ISSUE #232: Assisted exercises treat heavier assistance as PR — lower weight should be the record
### What did you do?

1. Created a routine with Assisted pull-up (leverage machine, id 0017).
2. Logged a workout with 30 kg assistance × 8 reps and finished it.
3. Logged a second workout with 20 kg assistance × 8 reps and finished it.

With assisted machines, less assistance is harder, so 20 kg should be the new Best / PR and be shown as Previous best: 30 kg — new record! Next time should suggest less assistance.

### What happened instead?

20 kg was not recognized as a PR — 30 kg remained as Best. The app treats heavier assistance as better (as with normal lifts), so progression also suggests the wrong direction.

Self-hosted (docker compose)

### Browser & OS

Chrome 153 on Android

dpatrongomez:
Assisted machines (e.g. assisted pull-up, 0017) reduce load, so less assistance is harder and should be the record. Previously every comparison used max, marking heavier assistance as PR.

- add isAssisted() helper in exercises.js (name contains assist/asist or explicit assisted flag)
- history.js: bestWeightForEntry/bestWeightFor return min for assisted, add hasWeightHistory/isWeightPR
- exercise-history.js: track min best for assisted
- onerm.js: invert bestSetOf/best1RM/is1RMRecord for assisted
- progression.js: invert readSession weight (min) and prescription direction (hit = less assistance, stall = more), skip Epley deload for assisted
- sheets.jsx: use isWeightPR and min/max for exWeights, prevBest and topWeight sheet
- Workout.jsx: compute best with min for assisted
- Stats.jsx: track min exBest for assisted
- sync-merge.js/import-csv.js: merge min for assisted
- progression.test.js: exclude assisted from generic LIFT fixture

Fixes reported bug: assisted pull-up with 20kg should be PR over 30kg.

Verification: frontend npm test 105/105, 1468/1468 passed.

Captura de pantalla 2026-09-16 a las 13 11 33
Captura de pantalla 2026-09-16 a las 13 11 51

Closes #232


DuarteSantos8:
Thank you for this, and sorry it took a while. The bug is real and it shipped fixed in **v1.3.8** — but with a different detector, so I am closing this rather than merging it.

The reason is `isAssisted()`. Matching the exercise name alone catches **29** entries in the catalogue, and only **8** of them are the machine. The other 21 are partner-assisted stretches (`assisted lying glutes stretch`), a medicine-ball twist (`assisted motion russian twist`), band and bodyweight leg curls (`band assisted pull-up`, `self assisted inverse leg curl`). On those the weight is ordinary load — heavier is harder — so inverting them would have been the same bug pointed the other way, and quieter, because nobody would think to check a stretch.

What shipped uses equipment *and* name: the leverage machine whose name says assisted, which is exactly those 8. `assisted: true` (or `false`) on a custom exercise or a routine's config overrides it either way, which covers anything the catalogue does not know — that part is your idea and it is in.

Beyond your list it also covers the sync merge (which kept the *larger* kept weight and handed the help back on the next sync) and the 1RM calculator in the detail sheet (it was offering a one-rep max of the assistance). Tests are in `frontend/src/lib/assisted.test.js`.

One thing your PR surfaced that I would not have found: the progression suite picked its "plain lift" fixture as the first non-cardio, non-leg, non-bodyweight exercise in the catalogue — which is `assisted chest dip`. It had been testing normal progression against an assistance machine all along.

PR #241, opened 2026-09-17
Collapse completed workout exercises without scroll jumps

helgehelge123:
## Summary

Long workout lists keep all set editors visible after an exercise is finished. Add **Workout view → Layout → Collapse completed exercises** for List and Compact views so finished exercises become a small name-and-set-count summary when the user moves to another exercise.

- The option is off by default and belongs to the running session, like the existing layout override. The current exercise stays open; **Set current** reopens a finished exercise for corrections.
- A superset collapses as one group only when every set of every member is complete. Empty exercises, unfinished warm-ups and partially completed unilateral sets remain expanded. Logged weights, reps and completion data are unchanged.
- Capture a surviving visible tile before the DOM changes and restore its screen position afterward. Account for the Layout sheet's body scroll lock and delayed keyboard restoration. At either document boundary, reserve only the space needed to keep that tile in place; later expansion reclaims it.
- Add translations in all 14 locales and regression coverage for completion, reopening, supersets, repeated exercises, scroll boundaries and modal restoration.

## Validation

- Node 22: **106 frontend test files, 1,486 tests passed**.
- Production frontend build passed; existing bundle-size/dynamic-import warnings remain.
- Locale parity: 14 locales, 1,293 keys each. Source-string check passed.
- Fatigue probe: 108,000 monotonicity and 14,076 history-edit comparisons passed.
- Clicked through a local guest workout in the Codex in-app browser and Google Chrome, using synthetic exercise data.

| Browser | List viewports (CSS px) | Compact viewports (CSS px) |
| --- | --- | --- |
| Codex in-app browser | 320×568, 390×844, 844×390, 768×1024, 1280×600, 1440×900 | 390×844, 844×390, 1440×900 |
| Google Chrome | 390×844, 1280×600 | 844×390, 1440×900 |

Measured the surviving tile's bounding rectangle before and after real coordinate clicks (locator auto-centering otherwise changes the starting scroll position). Final checks stayed within one CSS pixel with no horizontal overflow. Also checked completing the current exercise, switching, reopening, and toggling the option through the Layout sheet while scrolled. Safari, Firefox and physical mobile devices were not available for this manual test.


kurktchiev:
Heads-up on a shared hunk: #257 changes the same cleanup in `Modals.jsx` (the un-pin and `scrollTo` when the last sheet closes), so whichever of the two lands second needs a rebase.

What #257 does on those lines: the pin/un-pin effect becomes a `useLayoutEffect`, so the un-pin and the `scrollTo(0, y)` land in the same frame as the sheet's unmount (iOS painted a frame at scroll 0 between the two, a visible jump on every RIR pick), and the restore is issued once more on the next animation frame because iOS applies scrolls asynchronously. It also makes `App.jsx` ignore the POP that a sheet's own history entry produces, so that path no longer adds a second scroll.

If this PR lands first, I rebase #257 and the animation-frame re-assert reads `restoreY` instead of `y`. If #257 lands first, `restoreY` here should feed both re-asserts (the animation-frame one and the 350 ms one), and the `workout-scroll-anchor` cancel should cover the animation-frame one too, or the anchor's new position is overwritten one frame later.

This is only about the shared lines, not a review of the collapse feature.

PR #243, opened 2026-09-17
iOS: the keyboard no longer closes the moment a field in a sheet is tapped (#242)

### LINKED ISSUE #242: [Safari] Keyboard glitching on input fields
### What did you do?

1. Go to the login page of the self hosted instance using safari on iPhone or iPad.
2. Tap "Create new profile".
3. Try to enter name and invite code.
4. See that the keyboard is opening and closing instantly.

The keyboard should stay open and I should be able to use the input fields.

### What happened instead?

The keyboard closes instantly. I am unable to enter a name or the invite code. Once in ~5 tries, the keyboard stays open. See attached recording.

https://github.com/user-attachments/assets/13988a59-f7ca-499c-a097-4a9e0e546568

Self-hosted (docker compose)

### Browser & OS

iPadOS 16.3.1 and Safari, known to also be a problem on more modern iOS devices with up-to-date OS

FabianReitz:
Fixes #242

## Cause

`lib/viewport-guard.js` blurs a focused text field when the page is displaced and no keyboard is open, so iOS puts the viewports back after the keyboard goes away. Tapping a field makes iOS scroll the visual viewport to it *before* the viewport shrinks for the keyboard. When that scroll event arrived first, the guard saw "no keyboard, page displaced, field focused" and blurred the field — the keyboard closed the moment it opened. Whether the scroll or the resize event lands first varies, which is why it worked about one time in five. The "Create new profile" sheet on the login screen is where it showed most.

## Change

- `realign()` takes a `release` field and only blurs that one. The resize path passes the field that held focus when the keyboard was seen closing, so the original case (tap the set's tick with the weight field focused) still lets go of it.
- The scroll path no longer blurs immediately. A field on a displaced page is let go only if it is still focused and no keyboard has reported itself open 700 ms later, so a keyboard that is never detected (floating keyboard, a close the resize event missed) is still covered.

## Testing

- `lib/viewport-guard.test.js`: four new tests (keyboard coming up, focus moving to the next field, only the named field is blurred, release when no keyboard ever shows); they fail on `main` and pass here. Full frontend suite passes.
- iPad, iPadOS 16.3.1, Safari against the Vite dev server over HTTPS: typing into the name and invite-code fields of "Create new profile", and the exercise picker search.


FabianReitz:
Here is a preview of the fix. Compare with the screen recording on #242:


https://github.com/user-attachments/assets/eef01c17-486b-4acb-a3e8-a3239fdef69d

DuarteSantos8:
Thank you for this — the diagnosis in it is right, and the fix that shipped in **v1.3.8** is built on your report and your reading of `realign()`.

I went with the simpler invariant @hectorroaz proposed in #242 instead, so I am closing this rather than merging it. The reason is in their instrumented numbers: on a device where the browser chrome collapses while the keyboard animates in (Chrome for iOS, Safari with an auto-hiding toolbar), `innerHeight` moves *with* the visual viewport, so `keyboardOpen()` reports closed mid-animation and the **resize** path blurs the field too — not only the scroll path this PR guards. Your testing against the Vite dev server in a Safari tab could not surface that, because `innerHeight` is stable there.

What shipped: `realign()` does nothing at all while a text field has focus — no blur, no scroll, no unpin — which makes the unreliable detector harmless. That let the `release` / `OPEN_MS` plumbing go entirely.

You found this, wrote it up with a recording, and sent a fix within a day. That is exactly the report I want to get. Sorry to close it in favour of another shape.

PR #254, opened 2026-09-19
Exercise-name packs for it/de/es/fr plus two per-language English-name toggles

### LINKED ISSUE #253: [feat] Exercise-name packs for it/de/es/fr plus a per-language "English name in parentheses" toggle
### What problem does this solve?

openGym ships a catalogue of 1,324 built-in exercises with English names. Users who run the
app in Italian, German, Spanish or French (the three most-requested languages after the existing pt-BR,
hu, it packs) still see every exercise labelled in English — acceptable for a logbook, but the
rest of the UI is fully translated, and translated exercise names make workouts, plans and
stats immediately readable without having to mentally map foreign labels.

Separately, the translated exercise names are always followed by the English name in
parentheses. That is helpful while learning the local term, but it is noise for an experienced
user who wants a clean, fully-localized list. Today there is no way to turn it off, and no way
to express the choice *per language*.

### Proposed approach

- Add complete exercise-name packs for German, Spanish, French and Italian by translating the
  full 1,324-exercise catalogue. Each pack is authored as a JSON source file and generated by a
  one-off build script that fail-fast on unknown ids, empty values or an incomplete count, with
  a unit test asserting full catalogue coverage plus per-language translation rules.
- Existing languages without a pack (de-CH is a derived locale and inherits German) keep the
  current behaviour: English names, no toggle.
- Make the "(English name)" parenthetical a per-language user setting:
  - a persisted per-language map in the client store, defaulting to *shown* so existing
    profiles keep today's rendering;
  - Settings shows a single adaptive switch that reflects the active language (only when that
    language actually has a translated pack, so the option is hidden for languages that cannot
    use it) and remembers a separate choice for each language;
  - the app re-applies the exercise-name rendering when the setting changes.

### Dependencies

- [x] This doesn't require adding a new dependency

giulioleuci:
Closes #253

## Summary

Adds complete exercise-name packs for **Italian (it), German (de), Spanish (es) and French (fr)** — each translating the full 1,324-exercise built-in catalogue — plus **two independent per-language toggles** in Settings:

- **"English exercise names"** — the "(English)" parenthetical that follows translated names, now a per-language setting (previously always shown for the newly added languages).
- **"English names only"** — replaces the translated name with the original English one; while it is on, the parenthetical toggle is disabled for that language since the English name is already the displayed one.

Languages that already had translated packs (`pt-BR`, `hu`) are unaffected; each language remembers its own choice; `de-CH` keeps inheriting the German pack through the existing `derivePack` transform (ß→ss), so no extra pack to maintain.

## What changes

1. **Exercise-name packs** — every pack is generated by a build script from an authored JSON source (`scripts/exercise-name-sources/*.json` → `scripts/build-*-exercise-names.mjs` → `frontend/src/exercise-names/*.js`). The scripts fail hard on unknown ids, empty values or an incomplete count. Each pack ships a unit test asserting full catalogue coverage plus per-language translation rules (barbell→Langhantel/barra/barre/... dumbbell→Kurzhantel/mancuerna/haltère/... etc.).

2. **Per-language "(English name)" toggle** (`#253`, point 2) — the English parenthetical that follows translated exercise names is now a per-language user setting:
   - persisted per-language map (`DEF.enParens = {}`) in the store, defaulting to *shown* so existing profiles keep today's rendering;
   - Settings shows a **single adaptive switch** bound to the active language, remembered separately for each language;
   - the toggle row is **hidden for languages without an exercise pack**, so the option never appears for languages that cannot use it;
   - the app re-applies the exercise-name rendering when the setting changes.

3. **New "English names only" per-language toggle** — drops the translation entirely and shows the original English name. Since the parenthetical is redundant in this mode, the "English exercise names" switch is disabled (greyed out) while it is active. The two labels are translated in all 14 packs.

4. **i18n** — the new Settings labels are translated in all 14 locale packs (including a Brazilian-Portuguese override), with a `locale-coverage.test.js` asserting the keys exist and are non-blank in every shipped pack.

## Notes on translations

The exercise names in the `it`/`de`/`es`/`fr` packs were translated from English by an LLM (**OpenCode Zen**), then reviewed against the exercise catalogue. Please give them a native-speaker pass before relying on them in production.

## Tests

Full frontend suite passes: 131 files / 1,611 tests (same Node 22 runtime as the CI gate). No new dependencies.

## Issue

Fixes #253 — exercise-name packs for `it/de/es/fr` plus the per-language "English name in parentheses" toggle and a companion "English names only" toggle.


giulioleuci:
Implement #253

PR #265, opened 2026-09-20
Russian: the demo-build strings

dima-po:
The demo build (`VITE_DEMO=1`) brought its own copy — the seeded example profile, the Coach walkthrough in `lib/coach-demo.js`, the reset-demo-data controls in Settings and the self-hosting call to action in Login — but none of those strings reached a locale pack.

All fourteen packs are missing the same 33 keys, so the published demo is the one place a visitor meets openGym untranslated, whatever language they pick. This PR fills them in for Russian only; the other thirteen still need the same list, and I am happy to open this as an issue instead if you would rather do them in one sweep the way 631d01b did for the combine-routines strings.

Conventions followed from the existing pack: Coach is **Тренер**, second-person plural, `ё` written out. Demo-coach copy keeps the register of the real Coach strings so the walkthrough does not read like a different app.

`npm test` in `frontend/` passes — 105 files, 1468 tests.


dima-po:
Conflict note: #254 also appends to `frontend/src/locales/ru.js` (the two toggle strings), so this PR and that one will collide at the end of the file. Trivial to resolve — both are additions — and I am happy to rebase whichever way round you merge them.

